PR #4, opened 2018-02-14
Inception Export trained Checkpionts example

yupbank:
address https://github.com/kubeflow/kubeflow/issues/229

The goal is actually moving the whole piece  https://github.com/kubeflow/kubeflow/blob/master/components/k8s-model-server/README.md into here

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/4)

### DIFF
--- a/inception/Makefile
+++ b/inception/Makefile
@@ -1,0 +1,8 @@
+all: download dep export
+
+download:
+	./download_v3_checkpoint.sh
+dep:
+	pip install -r requirements.txt
+export:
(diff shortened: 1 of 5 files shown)


jlewi:
Per comment in that issue; https://github.com/kubeflow/kubeflow/issues/229#issuecomment-364619248

Can you provide more detail about the end goal? e.g. what will a final complete example look like?

jlewi:
@yupbank Ping?

yupbank:
Thanks... not it's not blocking, i will pick it up

jlewi:
Going to close this as stale.

PR #20, opened 2018-02-27
Create a end-to-end kubeflow example using seq2seq model (1/n)

ankushagarwal:
The goal is to have an end-to-end example of training and serving a non-trivial model using kubeflow.

This PR is mostly based upon kubeflow/README.md with changes tailored to this example.

Will incrementally check-in the other parts.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/20)

### DIFF
--- a/issue_summarization_github_issues/setup_a_kubeflow_cluster.md
+++ b/issue_summarization_github_issues/setup_a_kubeflow_cluster.md
@@ -1,0 +1,66 @@
+# Setup Kubeflow
+
+In this part, you will setup kubeflow on an existing kubernetes cluster.
+
+## Requirements
+
+*   A kubernetes cluster
+*   `kubectl` CLI pointing to the kubernetes cluster
+    *   Make sure that you can run `kubectl get nodes` from your terminal
+        successfully
+*   The ksonnet CLI: [ks](https://ksonnet.io/#get-started)
+
+Refer to the [user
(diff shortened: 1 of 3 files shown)


ankushagarwal:
@jlewi Can you review?

issue_summarization_github_issues/setup_a_kubeflow_cluster.md
```diff
@@ -0,0 +1,174 @@
+# Setup Kubeflow
+
+In this part, you will setup kubeflow on an existing kubernetes cluster.
+
+## Requirements
```
jlewi: Can we just link to the Kubeflow user guide as opposed to duplicating setup instructions here?

issue_summarization_github_issues/setup_a_kubeflow_cluster.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+If using GKE, we can configure our cloud environment to use GCP features with a
+single parameter:
+
+```
+ks param set kubeflow-core cloud gke --env=${KSONNET_ENV}
+```
+
+### Provision storage for training data
```
jlewi: See kubeflow/kubeflow#240 and kubeflow/kubeflow#241. Can we create docs in a central location that cover topics like how to setup storage?

@nkashy1 Were you going to take a look at doing that?

ankushagarwal:
I've updated the PR so that it references kubeflow user guide

PR #18, opened 2018-02-26
Skeleton testing framework

texasmichelle:
First stab at adding tests to this repo.

@jlewi what's the best way to see if this runs before merging? How can I run it locally?

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/18)

PR #22, opened 2018-02-28
Add a .gitignore

ankushagarwal:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/22)


texasmichelle:
Could you add a few more lines? I used the .gitignore from kubeflow/kubeflow as a starting point and we might not need everything in there, but it's a bit more robust. This is the one I've been using in a different PR: https://github.com/texasmichelle/examples/blob/master/.gitignore

ankushagarwal:
Added

ankushagarwal:
@texasmichelle Can you merge this? I don't think I have the permissions.

texasmichelle:
Ignoring unrelated presubmit tests since they were be resolved by [PR-18](https://github.com/kubeflow/examples/pull/18).

Issue #11, opened 2018-02-21
[GH Issue Summarization] Create a model server


texasmichelle:
Create a model server using TFServing.

Component of #14.

jlewi:
/assign @ankushagarwal 

Ankush can you describe the problems you were running into turning the model into a model that can be served with TF serving?

Would it be easier to serve the model using Seldon?

ankushagarwal:
The model used for issue summarization is very different from the examples that we've been using. For our image models, the model prediction looks something like this: `output = model(input)`

But for the issue summarization model, it looks something like this
```
output = '<START>'

intermediate_result = encoder_model(input)

while True:
  intermediate_result, next_char = decoder_model(intermediate_result, output)
  if next_char == '<STOP>':
    return output
  output += next_char

```

The first issue that I had was - exporting Keras models as Tensorflow models which can be used by TFServing - this is mostly done.

The second challenge that I have is understanding how TFServing works with 
1. multiple models
1. models with multiple inputs and multiple outputs


> Would it be easier to serve the model using Seldon?
I am not familiar enough with Seldon...

ankushagarwal:
I am having issues with the model exported from keras imported into tfserving. 

I get this error when I send a Prediction request to the TFServing server

```
AbortionError: AbortionError(code=StatusCode.INVALID_ARGUMENT, details="Expected multiples argument to be a vector of length 3 but got length 2
[[Node: Encoder-Last-GRU_1/Tile = Tile[T=DT_FLOAT, Tmultiples=DT_INT32, _device="/job:localhost/replica:0/task:0/device:CPU:0"](Encoder-Last-GRU_1/ExpandDims, Encoder-Last-GRU_1/Tile/multiples)]]")
```

Could not find a workaround for this. Will give seldon or tornado a shot to serve this Keras model. 

We can probably illustrate serving a model with TFServing in another example which trains a tensorflow model directly.

jlewi:
@cliveseldon @gsunner Do you think we should try to use Seldon here? 

Could we use the existing Seldon model server rather than creating our own Tornado stub?

Should we deploy the model using Seldon Core rather than deploying it directly with K8s resources?

cliveseldon:
@jlewi  For a sklearn model, seldon-core would seem to be a good choice. 

[...]

PR #36, opened 2018-03-09
Move from a custom tornado server to a seldon-core server for serving the model

ankushagarwal:
* seldon-core allows wrapping arbitrary models into a REST/GRPC API. We can leverage that feature instead of writing our own boilerplate code for an API server

Related to #11

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/36)


github_issue_summarization/notebooks/requirements.txt
```diff
@@ -0,0 +1,11 @@
+numpy
```
Jimexist: if you don't mind please use pipenv

ankushagarwal: Wrapping a seldon-core microservice requires a `requirements.txt` file: https://github.com/SeldonIO/seldon-core/blob/master/docs/wrappers/python.md#create-a-model-folder

github_issue_summarization/serving_the_model.md
```diff
@@ -1,21 +1,62 @@
...
+* `seq2seq_model_tutorial.h5` - the keras model
+* `body_pp.dpkl` - the serialized body preprocessor
+* `title_pp.dpkl` - the serialized title preprocessor
+
+# Wrap the model into a seldon-core microservice
+
+cd into the notebooks directory and run the following docker command. This will create a build/ directory.
+
+```
+cd notebooks/
```
jlewi: Does Seldon's model wrapper support reading the model from GCS/S3?

Could we run the python wrapper as a job on K8s? Could we create a ksonnet template to make it easy to launch such a job.

I don't think we should try to address either of those issues in this PR. But might be good to open up issues to track them.

/cc @cliveseldon @gsunner @ahousley

github_issue_summarization/serving_the_model.md
```diff
@@ -1,21 +1,62 @@
...
+```
+
+The build/ directory contains all the necessary files to build the seldon-core microservice image
+
+```
+cd build/
+./build_image.sh
+```
+
+Now you should see an image named `seldonio/issuesummarization:0.1` in your docker images. To test the model, you can run it locally using
```
jlewi: Can we tag it into a repo like GCR/DockerHub/Quay? We'll need that to deploy the model on K8s.

github_issue_summarization/serving_the_model.md
```diff
@@ -1,21 +1,62 @@
 # Serving the model
 
-We are going to use a simple tornado server to serve the model. The [server.py](notebooks/server.py) contains the server code.
+We are going to use [seldon-core](https://github.com/SeldonIO/seldon-core) to serve the model. [IssueSummatization.py](notebooks/IssueSummatization.py) contains the code for this model. We will wrap this class into a seldon-core microservice which we can then deploy as a REST or GRPC API server.
 
-Start the server using `python server.py --port=8888`.
+> The model is written in Keras and when exported as a TensorFlow model seems to be incompatible with TensorFlow Serving. So we're using seldon-core to serve this model since seldon-core allows you to serve any arbitrary model. More details [here](https://github.com/kubeflow/examples/issues/11#issuecomment-371005885).
```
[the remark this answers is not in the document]
ankushagarwal: Done https://github.com/kubeflow/examples/issues/38

github_issue_summarization/serving_the_model.md
ankushagarwal: > Does Seldon's model wrapper support reading the model from GCS/S3?

Not that I am aware of. Looking at the [instructions](https://github.com/SeldonIO/seldon-core/blob/master/docs/wrappers/python.md#create-a-model-folder), it seems to require all the files in a single folder which contains the code and the model to build the serving image.

Created an issue to track the feature for serving a seldon-core model: https://github.com/kubeflow/kubeflow/issues/405

github_issue_summarization/serving_the_model.md
ankushagarwal: Updated to use a GCR tag.

github_issue_summarization/serving_the_model.md
ankushagarwal: Looks like there is a seldon [component](https://github.com/kubeflow/kubeflow/tree/master/kubeflow/seldon) in kubeflow. In the next PR, I'll try to deploy this image on k8s using this component

github_issue_summarization/serving_the_model.md
```diff
@@ -1,21 +1,62 @@
...
 
-> The model is written in Keras and when exported as a TensorFlow model seems to be incompatible with TensorFlow Serving. So we're using our own webserver to serve this model. More details [here](https://github.com/kubeflow/examples/issues/11#issuecomment-371005885).
+# Prerequisites
 
-## Sample request
+Ensure that you have the following files from the [training](training_the_model.md) step in your `notebooks` directory:
```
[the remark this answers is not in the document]
jlewi: I guess if they ran the TFJOb on GCS they could just do gsutil to copy from the bucket to their pod.

github_issue_summarization/serving_the_model.md
cliveseldon: You can specify any persistent volume in definition of the SeldonDeployment in the componentSpec as its a standard Kubernetes PodTemplateSpec. So you can load your parameters from their rather than package them as part of the model image if you wish. The [kubeflow-seldon example](https://github.com/kubeflow/example-seldon) shows an example using an NFS volume to load the trained model parameters.

PR #45, opened 2018-03-12
[RL Agents] log and render to NFS, improve docs

cwbeitel:
This PR includes the following changes:
- Updates to the demo docs ([notebook](https://github.com/cwbeitel/examples/blob/50c9f8422d14fce4913fda067fca8f0fa76c3baf/agents/doc/demo.ipynb) and [readme](https://github.com/cwbeitel/examples/blob/50c9f8422d14fce4913fda067fca8f0fa76c3baf/agents/README.md)) which were out-dated in multiple places
- Removed unused tools/ dir
- Update main readme and contributing to reference the example
- A [minimalist test](https://github.com/cwbeitel/examples/blob/50c9f8422d14fce4913fda067fca8f0fa76c3baf/agents/trainer/task_test.py) that runs the trainer for 10 steps (that currently doesn't appear to be triggered by the presubmit robot)
- Inclusion of kubeflow vendor/ tf-job code, [see](https://github.com/cwbeitel/examples/tree/master/agents/app)
  - Optionally this could be replaced by setup logic that adds and pulls this dependency instead of checking it in.
- Illustrates how logging and rendering to an attached volume can simplify the process of viewing logs with TensorHub and exploring render outputs.
  - Storage in user-space allows it to be sym-linked into directory tree watched by TensorHub extension (which is running tensorboard --logdir=/home/jovyan)
  - Not specific to the use of NFS but rather to using attached volumes instead of storage provider client libs.
  - Alternatively could be accomplished with things like gcsfuse and s3-fuse but those seem less appropriate given we're running on Kubernetes

NFS specifically:
- Changes
  - The addition of NFS volume references and mounts to the train and render Ksonnet [prototypes](https://github.com/cwbeitel/examples/tree/50c9f8422d14fce4913fda067fca8f0fa76c3baf/agents/app/components).
  - References to this approach in notebook
- Benefits
  - Vendor agnosticism
  - Seems like a simplification of the example - people have to ks apply the NFS component instead of creating a gcloud service account and adding that as a cluster secret.
- Doesn't prevent use of GCS in the way illustrated in the previous example (copies using storage client lib)
- Perhaps in the future there should be logic for which type of volume to attach and a utility to check that NFS is deployed and if not link to docs.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/45)


jlewi:
/assign @texasmichelle 

@cwbeitel Please expand the PR description; please explain the motivation for the PR; i.e. why render to NFS?

cwbeitel:
@jlewi Expanded the description.

@texasmichelle Is there anything you're waiting on me for? Wondering if perhaps what I missed is that you're an approver on this and I needed to arrange / assign reviewers? If none of these there's no time pressure on reviewing this and it will become out-dated once the methods from https://github.com/kubeflow/examples/pull/60 are applied to this example.

cwbeitel:
Closing this PR, rebasing to recent linter-related changes, and initially simplifying down to only additions to restore example functionality (changing hparams_base to dict,  including vendor/) and minor cleanup (removing tools). Sending use of NFS and major changes to docs regarding this as separate PR.

PR #51, opened 2018-03-18
Update seq2seq_utils.py - Fixed call to `corpus_bleu`

hamelsmu:
Found a mistake with calculation of BLEU Score.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/51)


hamelsmu:
@jlewi  I fixed the convenience function that calculates BLEU Score.  Previous version resulted in the metric being calculated incorrectly (much, much worse than reality!)

PR #61, opened 2018-03-28
Add .pylintrc

texasmichelle:
Resolves failing presubmit tests.

Copied from [kubeflow/kubeflow](https://github.com/kubeflow/kubeflow/blob/master/.pylintrc).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/61)


agents/trainer/task.py
```diff
@@ -14,19 +14,16 @@
...
 import pprint
 import uuid
 
-import pip
-import tensorflow as tf
 from google.cloud import storage
+import tensorflow as tf
 
 import agents
-import pybullet_envs  # To make AntBulletEnv-v0 available.
```
cwbeitel: Hey removing 'pybullet_envs' will break the example. By importing it the environments get registered and are then available to the trainer. Just add a "#pylint: disable=unused-import" for that one. Good changes on the rest. I made some of these changes in https://github.com/kubeflow/examples/pull/45/files but didn't deal with the issues cased by that base hparams object. I'm thinking that should just be converted to a dict or an hparams object given it would be pretty unsightly to put an pylint ignore comment on each of those lines.

cwbeitel: And when I say "base hparams object" I'm referring to hparams_base().

agents/trainer/task.py
```diff
@@ -113,39 +110,39 @@ def hparams_base():
   """Base hparams tf/Agents PPO """
```
cwbeitel: Taking another look this will work fine if this is just converted into a dictionary for now.

agents/trainer/task.py
```diff
@@ -243,23 +240,22 @@ def gcs_upload(local_dir, gcs_out_dir):
     blob.upload_from_filename(local_file_path)
 
 
-def main(unused_argv):
```
cwbeitel: I would have thought main() would still need to be main(_) to work with tf.app.run but if this still works then cool.

agents/trainer/task.py
```diff
@@ -243,23 +240,22 @@ def gcs_upload(local_dir, gcs_out_dir):
...
+def main():
   """Run training."""
   tf.logging.set_verbosity(tf.logging.INFO)
 
   if FLAGS.debug:
     tf.logging.set_verbosity(tf.logging.DEBUG)
 
-  run_config = tf.contrib.learn.RunConfig()
```
cwbeitel: This can be removed if you have the inclination since the config is not being used and calling it doesn't have any side effect.

agents/trainer/task.py
```diff
@@ -243,23 +240,22 @@ def gcs_upload(local_dir, gcs_out_dir):
...
+  tf.contrib.learn.RunConfig()
 
   log_dir = FLAGS.logdir
 
-  agents_config = _get_agents_configuration(
-      FLAGS.hparam_set_id, log_dir, run_config.is_chief)
```
cwbeitel: Good call yeah _get_agents_configuration is no longer getting the hparam object dynamically from globals() so this simplification is good.

agents/trainer/task.py
texasmichelle: Replaced this with main(_)

texasmichelle:
Thanks for the quick review @cwbeitel!

cwbeitel:
Any time!

-CB

texasmichelle:
Could you add an lgtm so that it auto-merges?

cwbeitel:
Oh I didn't realize I was an actual reviewer on this I was just commenting. Let me pull and try it out real quick.

cwbeitel:
/lgtm
Tried it out and training fails because a default it's expecting from hparams_base isn't available so as I was saying turning this into a dict instead of commenting it out will keep the defaults and prevent the linter errors but that's something for me to fix in another PR. Approving this so it's not causing errors in tests of other examples which is the point here.

texasmichelle:
Thanks - it will be good to unblock the other PRs

ankushagarwal:
@texasmichelle : This looks great, thanks for doing this!

Since a lot of contributors are expected to contribute to examples in python, it will be good to have instructions in the repo to be able to run pylint locally especially since the Details link provided by the kubeflow-examples-presubmit job only lists the python files with lint issues and does not actually print out the issues. Also do you use something like `autopep8` to automatically fix the lint issues or do you manually fix them?

texasmichelle:
That's a great idea to add instructions. Let's merge this to unblock the other outstanding PR's and I'll create a new one with linting guidelines.

texasmichelle:
It took me a while to get versions straightened out & dig out the logs from the kubeflow-ci cluster, so it's worth documenting. I did these manually since there weren't many of them, but autopep8 is a good idea!

ankushagarwal:
@texasmichelle : FYI tf-operator docs suggest using yapf : https://github.com/kubeflow/tf-operator/blob/master/developer_guide.md#code-style

texasmichelle:
Good find - I added that link to the issue

PR #74, opened 2018-04-06
Added tutorial for object detection distributed training

ldcastell:
Added steps on how to leverage kubeflow tooling to
submit a distributed object detection training job
on an existing kubernetes cluster

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/74)


ldcastell:
Hello! Have you guys had the chance to look into this PR? is this aligned to the community goals? can I improve it?

ankushagarwal:
@ldcastell Thanks for your contribution! I skimmed over it and it looks good to me. Will take a detailed look at this on Monday. 

For the next steps, it will be great to serve it through TF Serving since the other example uses seldon-core for serving.

object_detection/distributed_training/setup.md
```diff
@@ -0,0 +1,51 @@
...
+```
+kubectl -n kubeflow get pods
+NAME                              READY     STATUS    RESTARTS   AGE
+ambassador-7987df44b9-4pht8       2/2       Running   0          1m
+ambassador-7987df44b9-dh5h6       2/2       Running   0          1m
+ambassador-7987df44b9-qrgsm       2/2       Running   0          1m
+tf-hub-0                          1/1       Running   0          1m
+tf-job-operator-78757955b-qkg7s   1/1       Running   0          1m
+```
+## Preparing the training data
```
ankushagarwal: I don't think that this is a good solution to write data to k8s worker nodes directly. When using k8s, we should not have to deal with nodes directly. There are a lot of problems associated with dealing with nodes directly. For example - we do not know which nodes will be scheduled for the tfjob worker pods. How do you cleanup this data? We might not even have ssh access to the nodes.

There are many different ways to do this. One way could be that you can create a preprocessing job which requests a PersistentVolumeClaim and stores data after preprocessing in the PVC.

The PV can then be mounted in `ReadOnlyMany` mode by the TFJob workers and use it for training.

ldcastell: Thanks for the feedback @ankushagarwal !!
I based the example on a very small cluster (2 nodes) so that's the reason behind those steps. I completely agree with you, this is a very specific case. I'll change the tutorial to use a more general way of getting and using the data.

ldcastell:
Hi @ankushagarwal, @texasmichelle 
I know there's a lot going on currently with kubeflow, but if you have some time I will gladly appreciate your feedback on this PR. Please let me know what you think and how can I make it better.

Thanks!

ankushagarwal:
Hi @ldcastell 
I'll try to review this end of the week.

ldcastell:
Hi @ankushagarwal 
Thanks for taking the time to review this patch :)

object_detection/setup.md
```diff
@@ -0,0 +1,59 @@
+## Setup Kubeflow
+### Requirements
+
+ - Kubernetes cluster
+ - Access to a working `kubectl` (Kubernetes CLI)
+ - Ksonnet CLI: [ks](https://ksonnet.io/)
+
+### Setup
+Refer to the [user guide](https://github.com/kubeflow/kubeflow/blob/master/user_guide.md) for instructions on how to setup kubeflow on your kubernetes cluster. Specifically, look at the section on [deploying kubeflow](https://github.com/kubeflow/kubeflow/blob/master/user_guide.md#deploy-kubeflow).
```
ankushagarwal: Update link to https://www.kubeflow.org/docs/about/user_guide/

ldcastell: Good catch! thanks @ankushagarwal!, I have updated the links to the new location of the user guide

Issue #91, opened 2018-04-24
[GH Issue] Use persistent volumes for the data


jlewi:
To support Katacoda kubeflow/examples#89 we should remove the need for GCP credentials.

1. For the output we should support using PD and making it easy for users to set the PD by parameters.
1. The input is trickier. I think GCS requires an account even for public buckets. If its a single file we could use the http URL to access it. 
 
    * My suggestion would be to create a script that would copy the data using curl to a PD. We could then  run that script as a K8s job.

texasmichelle:
Anyone still interested in pursuing this? Is it still relevant?

PR #97, opened 2018-04-25
Add a job to download the data to PVC.

jlewi:
* This is the first step to doing training and serving using a PV as opposed
  to GCS.

* This will make the sample easier to run anyhere and in particular on Katacoda.

* This currently would work as follows

User creates a PVC

ks apply ${ENV} -c data-pvc

User runs a K8s job to download the data to PVC

ks apply ${ENV} -c data-downloader

In subsequent PRs we will update the train and serve steps to load the
model from the PVC as opposed to GCS.

Related to #91

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/97)


ankushagarwal:
LGTMed too soon : `github_issue_summarization/ks-kubeflow/components/params.libsonnet` should probably also have `data-downloader`

@jlewi

PR #109, opened 2018-04-28
Improvements to the tensor2tensor trainer for the GitHub summarization example.

jlewi:
* Simplify the launcher; we can just pass through most command line arguments and not
  use environment variables and command line arguments.

  * This makes it easier to control the job just by setting the parameters in the template
    rather than having to rebuild the images.

* Add a Makefile to build the image.

* Replace the tensor2tensor jsonnet with a newer version of the jsonnet used with T2T.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/109)


github_issue_summarization/tensor2tensor/github/Dockerfile
```diff
@@ -1,14 +1,19 @@
-FROM gcr.io/kubeflow-images-public/tensorflow-1.7.0-notebook-gpu:latest
+# Docker image to train a model using T2T
+#
+# For GPU use gcr.io/kubeflow-images-public/tensorflow-1.7.0-notebook-gpu:latest
+# For CPU use gcr.io/kubeflow-images-public/tensorflow-1.7.0-notebook-cpu:latest
+ARG BASE_IMAGE=gcr.io/kubeflow-images-public/tensorflow-1.7.0-notebook-gpu:latest
```
ankushagarwal: We should pip install as jovyan user instead of root.

```
ARG BASE_IMAGE=gcr.io/kubeflow-images-public/tensorflow-1.7.0-notebook-gpu:latest
FROM $BASE_IMAGE

RUN pip install tensor2tensor h5py

USER root

RUN apt-get install -y jq

COPY __init__.py github/__init__.py
COPY github_problem.py github/github_problem.py
COPY t2t_launcher.sh github/t2t_launcher.sh
RUN chmod a+rx github/t2t_launcher.sh

RUN chown -R jovyan:users /home/jovyan/github

USER jovyan
```

github_issue_summarization/ks-kubeflow/components/tensor2tensor.libsonnet
```diff
@@ -1,95 +1,148 @@
...
+
+    local containerEnv = [
+      {
+        name: "PYTHONPATH",
+        value: "/home/jovyan",
+      }
+    ],
+    local baseCommand = [      
+      "/home/jovyan/github/t2t_launcher.sh",
+      "--train_steps=" + std.toString(updatedParams.trainSteps),
```
ankushagarwal: nit: `std.toString` is redundant and can be removed from all places.

> String concatenation will implicitly convert one of the values to a string if necessary

http://jsonnet.org/ref/spec.html

github_issue_summarization/tensor2tensor/github/Dockerfile
jlewi: Done.

jlewi:
Thanks. PTAL.

PR #122, opened 2018-06-01
Fix utf-8 encoding issues in data pre-processing

activatedgeek:
This update fixes the occasional `utf-8` errors during tokenization.

A representative error looked like

```
RuntimeError: UnicodeDecodeError: 'ascii' codec can't decode byte 0xe6 in position 654: ordinal not in range(128) [while running 'BigQueryGithubFiles/Extract Function Info']
```

This was happening at the Extract Function Info phase which needs to store the tokenized values into a Python `dict`. The assumption for ignoring such characters is that they would not add much value to the language model.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/122)


ankushagarwal:
/lgtm

Can you add a short description of what the problem was in the PR description?

PR #125, opened 2018-06-05
Add num workers options to Dataflow

activatedgeek:
This adds ability to mention number of Dataflow workers and the machine type.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/125)


code_search/preprocess/scripts/process_github_archive.py
```diff
@@ -7,13 +7,18 @@
 
 def parse_arguments(args):
   parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-  parser.add_argument('-i', '--input', metavar='', help='Path to BigQuery SQL script')
-  parser.add_argument('-o', '--output', metavar='',
+  parser.add_argument('-i', '--input', metavar='', type=str, help='Path to BigQuery SQL script')
```
jlewi: Do wee need to set metavar?

activatedgeek: I generally do that to have a clean help print (`-h`). Otherwise, those variables in caps interfere with readability.

Issue #121, opened 2018-05-30
Semantic Code Search Example


activatedgeek:
This issue tracks the discussion for the Semantic Code Search example.

# Objective

The aim is to implement an end-to-end Semantic Code Search pipeline and deploy on top of Kubeflow.

# Pipeline

- [x] Data Ingestion and Preprocessing from BigQuery via Dataflow (#120, #122, #125, #128)
- [ ] Train/Test/Validation Data Split
- [x] (Single Machine) Tensor2Tensor Datagen and Trainer (#127)
- [x] (Single Machine) Language Modeling for docstrings (#129)
- [x] Training with Kubeflow (#131)
- [ ] Export models for reuse
- [ ] Data Indexing for search
- [ ] Model Serving for queries

## Enhancements
- [ ] Distributed Training
- [ ] Store time taken by each step (and not rely on Dataflow times)
- [ ] Use TFTransform to generate Vocabs (behaves nicely with Dataflow)?
- [ ] Argo workflow to integrate all steps?

## Acknowlegements

This project derives from [hamelsmu/code_search](https://github.com/hamelsmu/code_search) and aims to reuse some parts from there (albeit some copy-paste code).

jlewi:
Should we split this up into separate issues and use a new label to keep track of them; e.g.
area/examples/code_search?

jlewi:
Created label:
[area/example/code_search](https://github.com/kubeflow/examples/labels/area%2Fexample%2Fcode_search)

activatedgeek:
I wanted a place to track the entirety of the code search example. I can move this to somewhere else if you prefer. I'm anyways sending out smaller PR chunks for isolated tasks/problems.

activatedgeek:
Closing this in favor of isolated issues.

/close

PR #129, opened 2018-06-06
Language modeling using Transformer Networks

activatedgeek:
This carries towards #121 and trains a language model. Note that integration with previous steps is still a TODO and outside the scope of this PR. Hyper-parameters might need playing around with and will also be taken up separately.

/cc @jlewi @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/129)


jlewi:
What are we using for the target text in this case?

activatedgeek:
In this, the target text is the  input text itself. The idea being that we want to embed the tokens in the context we see them.

PR #131, opened 2018-06-08
Isolate t2t execution into docker

activatedgeek:
This isolates the Tensor2Tensor training into a `Dockerfile` which can be directly deployed (for example into a Kubeflow TFJob).

It has a utility `run.sh` which abstracts out certain housekeeping stuff from the user.

/cc @ankushagarwal @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/131)

### DIFF
--- a/code_search/README.md
+++ b/code_search/README.md
@@ -72,7 +72,7 @@ $ python preprocess/scripts/process_github_archive.py -i files/select_github_arc
 ## 2. Model Training
 
 A `Dockerfile` based on Tensorflow is provided along which has all the dependencies for this part of the pipeline. 
-By default, the based off Tensorflow CPU 1.8.0 for `Python3` but can be overridden in the Docker image build using
+By default, it is based off Tensorflow CPU 1.8.0 for `Python3` but can be overridden in the Docker image build using
 the following command
 
 ```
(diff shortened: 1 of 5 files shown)


activatedgeek:
There seems to be an issue with writing the results back into the mounted volume.

code_search/README.md
```diff
@@ -68,47 +69,54 @@ $ python preprocess/scripts/process_github_archive.py -i files/select_github_arc
          --max-num-workers 16
 ```
 
-## 2. Function Summarizer
+## 2. Model Training
 
-This part generates a model to summarize functions into docstrings using the data generated in previous
-step. It uses `tensor2tensor`.
+A `Dockerfile` based on Tensorflow is provided along which has all the dependencies for this part of the pipeline. 
+By default, the based off Tensorflow CPU 1.8.0 for `Python3` but can be overridden in the Docker image build using
```
jlewi: nit: Grammer error "By default, the based off"

activatedgeek: Thanks for catching that.

code_search/language_task/run.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
activatedgeek: Certainly, I will be adding the k8s (Kubeflow first) job next. I wanted to keep the scope of this PR limited only to the Docker image building.

PR #141, opened 2018-06-13
Add custom metrics, write raw tokens to GCS

activatedgeek:
This step is meant to be used as a step before generating the negative samples for the new architecture. Using GCS for output storage. The cost to write to GCS seems to be at par with the cost to write to BigQuery.

/cc @jlewi @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/141)


activatedgeek:
This PR makes the data ready to be consumed for further processing via a GCS link.

PR #143, opened 2018-06-15
Language task on kubeflow

activatedgeek:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/143)


activatedgeek:
cc #144.

activatedgeek:
The YAML generated here is representative but should be ready to use. The documentation will change (esp. the `--problem` argument) because currently a new architecture for the problem is in development.

activatedgeek:
@jlewi Oh I just wanted to put a representative example on how to use this `t2t-job` library. This will be replaced in another PR with the new architecture that we are working towards.

Issue #136, opened 2018-06-10
[code_search] Tensor2Tensor model to learn similarity code and search queries


jlewi:
We would like to use a single Tensor2Tensor model to learn how to map code and language describing the code to the same search space.

We should be able to do this using the transformer model in T2T.

/assign @activatedgeek 
/cc @hamelsmu

PR #146, opened 2018-06-18
Add a new similarity transformer model, register new problem

activatedgeek:
This is for #136.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/146)


activatedgeek:
This is just the skeleton but is needed to be a part of #152 .

PR #152, opened 2018-06-26
Add new TF-Serving component with sample task

### LINKED ISSUE #138: [code_search] TF Serving to encode queries
To encode search queries we will want to use TF Serving. This will mean

* Using the T2T exporter to export the model for serving
* Create a TF serving deployment

For more info see the [doc](https://docs.google.com/document/d/1lqXmL_32cmDQ8z7MSl3FJ6p45RF-KDw5isGLyxPXsa8/edit#heading=h.6bl634ymn0aw)

### LINKED ISSUE #151: [code_search] Ksonnet app for nmslib server
A Ksonnet app for nmslib server which can do the PVC provisioning and mounting.

/area example/code_search

activatedgeek:
- [x] TF Serving Ksonnet Component (Fixes #138 )
- [x] Add support for input encoding in NMSLib server (`tf.train.Example` format)
- [x] Add REST API port flag to TF Model Server binary in container spec
- [x] NMSLib server Ksonnet Component (Fixes #151 )

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/152)


code_search/app/code_search/nmslib/search_engine.py
```diff
@@ -0,0 +1,69 @@
...
+    encoded_query = encode_query(encoder, query_str)
+    data = {"instances": [{"input": {"b64": encoded_query}}]}
+
+    response = requests.post(url=self._serving_url,
+                             headers={'content-type': 'application/json'},
+                             data=json.dumps(data))
+
+    result = response.json()
+    for prediction in result['predictions']:
+      prediction['outputs'] = decoder.decode(prediction['outputs'])
```
jlewi: Why are you decoding it? Isn't the point of the TF serving server to embed the the search query in the space? So doesn't the TF serving model return the embedded vector?

code_search/app/code_search/t2t/query.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def encode_query(encoder, query_str):
+  """Encode the input query string using encoder. This
+  might vary by problem but keeping generic as a reference.
+  Note that in T2T problems, the 'targets' key is needed
+  even though it is ignored during inference.
+  See tensorflow/tensor2tensor#868"""
+
+  encoded_str = encoder.encode(query_str) + [text_encoder.EOS_ID]
```
jlewi: If you're calling encode here, then what purpose is the model being served by TF Serving serving?

code_search/app/code_search/nmslib/cli.py
```diff
@@ -0,0 +1,82 @@
+import sys
```
jlewi: Add a doc string explaining what this CLI is for.

code_search/app/code_search/nmslib/search_engine.py
```diff
@@ -0,0 +1,69 @@
...
+import requests
+import nmslib
+import numpy as np
+from tensor2tensor import problems  # pylint: disable=unused-import
+from code_search.t2t.query import get_encoder_decoder, encode_query
+
+
+class CodeSearchEngine:
+  """This is a utility class which takes an nmslib
+  index file and a data file to return data from"""
```
jlewi: What is in the data file?

code_search/app/code_search/nmslib/cli.py
```diff
@@ -0,0 +1,82 @@
...
+
+def server():
+  args = parse_server_args(sys.argv[1:])
+
+  if not os.path.isdir(args.tmp_dir):
+    os.makedirs(args.tmp_dir, exist_ok=True)
+
+  # Download relevant files if needed
+  index_file = maybe_download_gcs_file(args.index_file, args.tmp_dir)
+  # data_file = maybe_download_gcs_file(args.data_file, args.tmp_dir)
```
jlewi: Delete commented out code.

code_search/app/code_search/nmslib/search_engine.py
activatedgeek: This part is only representative but data file is meant for a reverse lookup of human-readable code corresponding to the returned indices from nmslib NN search.

code_search/app/code_search/t2t/query.py
activatedgeek: Great point! I had similar concerns when starting but turns out TF Serving can itself not do anything with the incoming `instances` list, whether REST or gRPC, unless there is some part of the graph that does this transform. The sample model that I am using with this unfortunately doesn't have that part, so I had to manually do it. My next plan is to figure out the best way to write a T2T model which can handle all these things so that during export it also has such transformer functions. 

Consider this a work in progress but if nothing works out, this at least gets the inference done meanwhile.

code_search/app/code_search/nmslib/search_engine.py
activatedgeek: I wrote this part w.r.t another model and is representative of what "might" need to be done. From today, I am going to start investigating the graph (a simple single layer graph which takes input and gives some embedding out).

activatedgeek:
@jlewi This is ready for merge. I've added a few TODOs in the comments of the functions which might require a redo to make sure objectives align.

Issue #145, opened 2018-06-18
[GPU Serving] Serving example with GPUs? Inception?


jlewi:
There's lots of interest in serving with GPUs but we don't have a good example showing how to do this.

I think it would be nice to have one. 

A simple example might be inception with a simple front end that allows people to upload images for classification.

jlewi:
There's a pretty comprehensive tutorial here
https://github.com/google-aai/tf-serving-k8s-tutorial

Maybe we can just link to that or modify it to work with Kubeflow?

jlewi:
This [blog](https://cloud.google.com/blog/big-data/2017/09/performing-prediction-with-tensorflow-object-detection-models-on-google-cloud-machine-learning-engine) is a pretty good model.

It would be nice to make the following changes

- Create a web app to surface the results rather than just using a notebook
- Use GPUs for serving.

cliveseldon:
Using Seldon to serve should be straight forward as adding the requires resource requests for GPU resources in our CustomResource assuming you have set up the cluster to allow them. 

As you can use any Dockerized solution with Seldon you are free to compile your model with TensorRT or other frameworks.

jklaise:
> This blog is a pretty good model.
> 
> It would be nice to make the following changes
> 
> Create a web app to surface the results rather than just using a notebook
> Use GPUs for serving.

@jlewi I'm working on implementing this example with serving done by Seldon including the training on GCP part as covered [https://cloud.google.com/blog/big-data/2017/06/training-an-object-detector-using-cloud-machine-learning-engine](here). Currently having issues getting the training part working, think it might be a Tensorflow version issue.

jlewi:
@jklaise cool. Are you planning on using TFServing with Seldon?

jklaise:
@jlewi No we don't, the plan is to provide an example to export a GCP trained model to a format that can easily be dockerized for serving with Seldon.

cliveseldon:
The plan is to try and use that example blog to train the model and export to run on seldon. Once done, I assume we can work with you to train on kubeflow directly and finally deploy with GPUs on an appropriate cluster to get a good e2e example?

jlewi:
@lluunn We now have a GPU serving example right?

lluunn:
Yes.
I can't close issue in examples repo.

PR #154, opened 2018-06-27
Example of TF Serving with GPU

lluunn:
#145

`python predict.py --utl=XX` should work.

Only files need review: README.md and predict.py
visualization_utils.py and standard_fields.py are copied from 
https://github.com/tensorflow/models/blob/master/research/object_detection/utils/visualization_utils.py

Next PR should move the python logic into a webapp and deploy it.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/154)


gpu_serving/object-detection-app/components/params.libsonnet
```diff
@@ -0,0 +1,16 @@
+{
+  global: {},
+  components: {
+    // Component-level parameters, defined initially from 'ks prototype use ...'
+    // Each object below should correspond to a component in the components/ directory
+    model1: {
+      cloud: 'gcp',
+      deployHttpProxy: true,
+      gcpCredentialSecretName: 'kai-sa',
```
jlewi: Why are you using the GCP secret kai-sa? I believe the GKE deploy secrets will create the secret user-sa does that work?

gpu_serving/predict.py
```diff
@@ -0,0 +1,46 @@
+import argparse
```
jlewi: This is a CLI for the example?

gpu_serving/object-detection-app/environments/kai4/globals.libsonnet
```diff
@@ -0,0 +1,2 @@
+{
```
jlewi: Lets not check in your environment.
I'd suggest adding it to the .gitignore file so it doesn't get checked in.

jlewi:
This is great.

@texasmichelle Any overall guidance about how we should be structuring the examples?

gpu_serving/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Prepare model
+Download the model from model zoo.
+The model should be in SavedModel format (including a `saved_model.pb` file and a
+optional `variables/` folder.
+
+```
+wget http://download.tensorflow.org/models/object_detection/faster_rcnn_nas_coco_2018_01_28.tar.gz
+tar -xzf faster_rcnn_nas_coco_2018_01_28.tar.gz
+gsutil cp faster_rcnn_nas_coco_2018_01_28/saved_model/saved_model.pb gs://YOUR_BUCKET/YOUR_MODEL/1/
```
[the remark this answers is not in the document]
lluunn: pushed

gpu_serving/object-detection-app/components/params.libsonnet
lluunn: I was reading  model in my bucket. user-sa should work.

gpu_serving/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+## Deploy serving component
+
+```
+ks init ks-app
```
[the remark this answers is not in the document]
lluunn: Done

gpu_serving/predict.py
lluunn: Yes, see README: `python predict.py --url=YOUR_KF_HOST/models/coco`

gpu_serving/object-detection-app/environments/kai4/globals.libsonnet
lluunn: Done.
But it's still in app.yaml. Not sure if there is a good way to avoid that.

lluunn:
@jlewi  Thanks, PTAL

gpu_serving/object-detection-app/components/params.libsonnet
jlewi: Can you change the default parameters to be parameters that will work for users here and below.

jlewi:
It looks like someone started checking in an object detection model here
https://github.com/kubeflow/examples/tree/master/object_detection

Is this the same example? Should we move it into that directory?

lluunn:
It's a different [model](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/running_pets.md), and it's more focus on how to do distributed training. 

I think we want to emphasize GPU serving here, the model being used is not that important. WDYT?

jlewi:
But both examples are doing object detection right?

I guess we can keep them separate but ideally we should have a single object detection example that can show both training and serving.

jlewi:
@lluunn Is there an issue to reconcile this with the pets example?

lluunn:
Filed #185

texasmichelle:
See comments on #185

jlewi:
The consensus on #185 seems to be to having a single object_detection model. 

Lunkai do you want to move these files into that directory? Maybe you replace what's in object_detection/tf-serving? And maybe combine the READMEs?

@texasmichelle does that work for you?

lluunn:
I think it's a little awkward to move this into object_detection now as that example currently uses many yaml and doesn't have a ks-app.
Should we either leave this PR here (or checked in for now) and then combine these two after we fix the issues in the object_detection/?

jlewi:
Why can't you make it a subdirectory of object_detection?

e.g. 

```
object_detection/tf_serving/ks-app
```

jlewi:
Do we even need the YAML file that's currently in object_detection/tf-serving? What is that file doing that your example isn't?

lluunn:
moved. I put the ks-app at the object_detection as I think other yaml files will be moved to the app too.

object_detection/tf_serving_gpu.md
```diff
@@ -0,0 +1,50 @@
+# Serving an object detection model with GPU
+
+## Setup
+
+If you followed previous steps to train the model, skip to deploy [section](#deploy-serving-component).
```
texasmichelle: The setup steps feel out of place here, since the title is Serving an object detection model with GPU. Could you integrate them with the other markdown files in this example?

texasmichelle:
What is `inputs.json` for? It is the last remaining file in gpu_serving and needs to be filed somewhere in object_detection if it is used for the example.

object_detection/tf_serving_gpu.md
lluunn: I think some users might only interested in trying out serving. This provides the steps to do if they skip the training part. And the first line is tell them to skip this if they've done training.

Does it make sense?

lluunn:
removed inputs.json. That's missed when moving.

jlewi:
Took another look. I think only significant comment I have is whether we should be telling users to use the ksonnet app provided in the example and just adding and environment to it rather than creating a new app.

My thinking is that based on other examples we will want to provide the app as an easy way of making a whole bunch of resources that are part of the example available to users.

lluunn:
Modify the instruction to tell user to cd to that ks app instead of ks init. PTAL, thanks

jlewi:
I had one minor comment abou the default value for the GCP secret name but other than that this LGTM.

lluunn:
Done. Thanks

jlewi:
/lgtm
/approve
/hold  for @texasmichelle 

@texasmichelle I know your busy with NEXT do you want @lluunn to wait until you can take another look?

PR #157, opened 2018-06-29
Update the GH summarization example to Kubeflow 0.2 and TFJob v1alpha2.

jlewi:
* Upgrade the ksonnet app to Kubeflow 0.2 rc.1
* Add the examples package.
* Add a .gitignore file and ignore all environments so that we won't pick
  up people's testing environments.

* Add tfjob-v1alpha2 component; this trains the model using Keras using
  TFJob v1alpha2.
  * Update the parameters so that we use the GCP secrets created as part
    of the Kubeflow deployment.

* Remove jlewi environment.

* Verified that training ran successfully and outputted a model to GCS

  * There was an error about some missing arguments to a logging statement
    but this can be ignored although it would be good to fix.

/assign @texasmichelle

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/157)


texasmichelle:
The unknown flag error is resolved by using tensor2tensor >= v1.6.3

/lgtm
/approve

PR #160, opened 2018-06-29
Fix model file upload

texasmichelle:
Update default params for tfjob-v1alpha2
Fix build directory path in Makefile

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/160)


github_issue_summarization/ks-kubeflow/components/params.libsonnet
```diff
@@ -55,15 +55,15 @@
       githubToken: "",
     },
     "tfjob-v1alpha2": {
-      name: "tfjob-v1alpha2",
-      image: "gcr.io/kubeflow-dev/tf-job-issue-summarization:v20180425-e79f888",
+      name: "tfjob-issue-summarization",
+      image: "gcr.io/kubeflow-examples/tf-job-issue-summarization:latest",
```
activatedgeek: Perhaps we should re-tag so as to not break this tomorrow when somebody updates the image?

jlewi: +1 Its best to pin to a specific image.

texasmichelle: Fair point - just pushed a new image & updated the tag

github_issue_summarization/ks-kubeflow/components/params.libsonnet
```diff
@@ -53,17 +53,18 @@
...
+      image: "gcr.io/kubeflow-examples/tf-job-issue-summarization:v20180629-v0.1-2-g98ed4b4-dirty-182929",
       input_data_gcs_bucket: "kubeflow-examples",
       input_data_gcs_path: "github-issue-summarization-data/github-issues.zip",
       output_model_gcs_bucket: "kubeflow-examples",
-      output_model_gcs_path: "github-issue-summarization-data/output_model.h5",
+      output_model_gcs_path: "github-issue-summarization-data",
       sample_size: "100000",
-      gcpSecretName: "user-gcp-sa",
-      gcpSecretFile: "user-gcp-sa.json",
+      gcpSecretName: "gcp-credentials",
```
jlewi: These aren't the right defaults. "user-gcp-sa" and "user-gcp-sa.json" are the defaults as set by our GKE deployment instructions.

PR #168, opened 2018-07-10
Search UI Components

### LINKED ISSUE #135: [code search] search web app
For the code search example we need a simple web app in which users can enter search queries and then see the results.

This issue covers adding a minimal web app. Follow on issues will handle various extensions like logging requests and responses.

Ref [doc](http://bit.ly/gh-kf-search)

activatedgeek:
Fixes #135 

screen shot 2018-07-10 at 5 29 41 pm

screen shot 2018-07-10 at 3 43 15 pm

/cc @jlewi @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/168)


activatedgeek:
**NOTE**: The results UI is only illustrative. The component is ready to be used for any given Github code. Once linked with the REST call, it will be automatically populated.

PR #183, opened 2018-07-17
Update pets-training to v1alpha2

royxue:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/183)

### DIFF
--- a/object_detection/jobs/pets-training.yaml
+++ b/object_detection/jobs/pets-training.yaml
@@ -1,77 +1,61 @@
 ---
-apiVersion: kubeflow.org/v1alpha1
+apiVersion: kubeflow.org/v1alpha2
 kind: TFJob
 metadata:
   name: pets-training
 spec:
-  replicaSpecs:
-  - replicas: 1
-    template:
-      spec:
-        containers:
-        - image: lcastell/pets_object_detection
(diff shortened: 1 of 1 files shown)


royxue:
@ldcastell Plz check this PR, which is with new apiVersion: v1alpha2, and be able to work.

object_detection/jobs/pets-training.yaml
```diff
@@ -1,77 +1,61 @@
...
+      replicas: 1
+      restartPolicy: OnFailure
+      template:
+        spec:
+          containers:
+          - name: tensorflow
+            image: lcastell/pets_object_detection
+            resources:
+              limits:
+                nvidia.com/gpu: 1
```
ldcastell: Question, if your cluster does not have GPU, will this make the job fail?

ldcastell: I agree with @jlewi I have started to look into parametrizing the manifests into a ksonnet app, and address this issue too.
I will see if I can push something by next weekend

royxue: Oh yes, I forget about this part, I gonna remove the GPU part.

and Agree with that ksonnet would be much better

PR #184, opened 2018-07-17
Integrate batch prediction

### LINKED ISSUE #180: [code_search] Batch Prediction of Similarity Transformer Model
We want to be able to embed the training data for indexing purposes and it is not really ideal to do that via lots of HTTP calls to TF Serving. It is better to have a batch inference routine which takes a lot of examples together.

/area example/code_search

activatedgeek:
This PR integrates Batch Prediction over the Tensorflow SavedModel as a Dataflow job. This needed a massive refactor because Dataflow needed the T2T model to be imported as a module to make the predictions. The unification of all modules was therefore necessary.

A bunch of changes:

- Migrate Python3 packages to Python2
- Updated README for Local Usage and exact translation to Kubeflow jobs
- Dataflow Job to generate string embeddings using Batch Prediction (Fixes #180)

A lot of the diffs are just because of the move, the main folders to be reviewed are

- `src/code_search/do_fns`
- `src/code_search/transforms`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/184)

### DIFF
--- a/code_search/src/setup.py
+++ b/code_search/src/setup.py
@@ -10,8 +10,10 @@ with open('requirements.txt', 'r') as f:
 
 CUSTOM_COMMANDS = [
   ['python', '-m', 'spacy', 'download', 'en'],
+  ##
   # TODO(sanyamkapoor): This isn't ideal but no other way for a seamless install right now.
-  # TODO(sanyamkapoor): Switch to upstream source after merge
+  # This currently uses a fork due to API limitations (See kubeflow/batch-predict#10). The
+  # API limitations have a workaround via kubeflow/batch-predict#9.
   ['pip', 'install', 'https://github.com/activatedgeek/batch-predict/tarball/fix-value-provider']
 ]
 


activatedgeek:
\uncc @wbuchwalter @zjj2wry

activatedgeek:
/cc @jlewi @ankushagarwal 

This PR is a relatively large refactor because virtually all the code requires the t2t modules. The Python2/3 split was causing maintainability issues, so I just merged into one python package with multiple modules.

activatedgeek:
> You should always use absolute import paths.
> Also you should import complete packages

Sure thing! Is there a reason we prefer absolute imports over relative? I spent some time reading about it but could not build a strong opinion on one or the other.

> What is code-search-predict? Is it a binary? I can't find it.

This is a python wrapper auto-generated after a `pip install` and is defined in the `setup.py`.

> Lets change this to use Kubernetes please. Lets not introduce docker.

All of the commands in README are currently for the user to run everything locally. I will certainly finalize the story and make it more accessible once I am finished.

activatedgeek:
> Can we make this an absolute import?
> Why is this import here? Is this to do with how DoFns get serialized?

Ah sorry, this slipped through. Yes, this has to do with the serialization of DoFns.

> Can you explain how this code works? Is next returning a single element?
> How can you have a yield statement when you aren't iterating over anything? Does it just return a single item and then a stop iteration on the next call

`next` takes in the CSV reader (which is an iterator) and gets the next value from the iterator. I am using the CSV Reader instead of manual string parsing because there were a lot of corner cases in how CSV uses the delimiter and the actual data characters. To iterate over the complete data, `next` needs to be called as many times as possible (until it throws an `StopIteration`). Here, I know that there's only ever a single data reference in the iterator and call it one time.

> What is meant by query friendly? Do you mean the strings to return to the user?

I think this needs a better doc. I'll update.

Issue #178, opened 2018-07-16
[Object Detection] Parameterize the K8s objects


jlewi:
The object detection example has a bunch of YAML manifests; e.g.
https://github.com/kubeflow/examples/blob/master/object_detection/jobs/pets-training.yaml

Can we turn these manifests into a ksonnet application and parameterize them so its easy for users to change things like the docker image, input/output location.

I believe ksonnet supports YAML now so we might not even need to convert the manifests to jsonnet.

ldcastell:
Hi @jlewi, sure I will start looking into it :)

jlewi:
The YAML functionality in ksonnet doesn't appear to be well documented.

I've had some success doing 

```
ks import -f path/to/yaml
```

I haven't figured out how to substitute parameters to YAML manifests. 

I have a question out to the ksonnet folks ksonnet/ksonnet#782

ldcastell:
@jlewi would it be something like this:
```
ks param set component metadata.name newname
```

PR #192, opened 2018-07-23
Parametrize Object detection example

ldcastell:
Added ksonnet prototypes to parametrize the object detection example 
Improved instructions 
Fixes: #178

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/192)


ldcastell:
@jlewi, @royxue I think this is ready for review. Thanks!

royxue:
@ldcastell Also, maybe we can consider merging the similar job into one? 
For example, to have only one get data job which create 3 pod at the same time, instead of `get-data`, `get-annotation`, `get-model`.

what do u think @jlewi

ldcastell:
Hi @royxue, 
Thank you for your feedback. I'm working on on-prem K8s that's why I added the step to create a PVC to store the data to a central place so all the worker pods can access it (perhaps we can add an statement into the steps to clarify why things are happening the way the are). 
I guess my intention with this example is to be as granular with the steps as possible. I'm pretty new to ML and I think having a clear understanding on what's going on step by step helps to get a better picture of the problem you're trying to solve. Having said that, I'm all for improving kubeflow's users experience through our examples so, we certainly can consolidate similar jobs into a single component like you said with the `get-data` jobs :).

royxue:
@ldcastell 
Yep, a `PVC` could not work without `PV`, if we wanna make the tutorial more smooth for the beginner level, I think we should add `create PV` together with `create PVC`.

something like this 
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: test-pv
spec:
  capacity:
    storage: 100Gi
  volumeMode: Block
  accessModes:
  - ReadWriteMany
  persistentVolumeReclaimPolicy: Delete
  storageClassName: local-storage
  local:
    path: /dataset/
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          - #some node name
---
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: local-storage
provisioner: kubernetes.io/no-provisioner
volumeBindingMode: WaitForFirstConsumer
---
# Add create PVC here
with storage name local-storage
```

ldcastell:
@royxue, I forgot to mention that I have a Ceph storage class configured in the K8s cluster and I can create PVC dynamically that way. 
I will update the description in the steps to clarify this situation and will add a new prototype to consolidate jobs.

royxue:
seems good to me

jlewi:
Thanks. Let me know when this is ready for another look.

ldcastell:
@jlewi I think this is ready for another look. Please let me know what you think :).

jlewi:
Thank you so much for doing this.

/lgtm
/approve

PR #197, opened 2018-07-24
[WIP] Refactor dataflow pipelines

activatedgeek:


activatedgeek:
/cc @ankushagarwal @jlewi 

This would probably be the last big PR. It fixes a lot of redundancies and makes the pipelines ready to be run locally.

PR #203, opened 2018-07-27
Add estimator example for github issues

### LINKED ISSUE #196: [gh_issue_summarization] distributed training using Keras
When we previously tried using distributed training with the Keras Seq2Seq model (#43)

We ran into the error in this [comment](https://github.com/kubeflow/examples/issues/43#issuecomment-380633578)

```
InvalidArgumentError (see above for traceback): Cannot colocate nodes 'Decoder-Word-Embedding/embeddings' and 'training/Nadam/gradients/Decoder-Word-Embedding/Gather_grad/Shape: Cannot merge devices with incompatible jobs: '/job:worker/task:0' and '/job:ps/task:0'
   [[Node: Decoder-Word-Embedding/embeddings = VariableV2[container="", dtype=DT_FLOAT, shape=[4278,300], shared_name="", _device="/job:ps/task:0"]()]]
```

using this code
https://gist.github.com/ankushagarwal/a2dab3aaf664a4296292c4ff330fb5e6

This is supposedly fixed with TensorFlow 1.9 and tf.Keras.

This would be worth retrying to see if we can make distributed training work with the Keras model.

inc0:
This is code input for doc about writing Keras for tfjob.

There are few todos:

1. bug in dataset injection, can't raise number of steps
2. intead of adding hostpath for data, we should have quick job + pvc
for this

Fixes #196

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/203)


jlewi:
Can we replace existing code  (e.g. delete some files) rather than creating yet another way to do training?

Did you get distributed training working with K8s?

jlewi:
Please update the PR description to provide more information about what's different about this PR from what currently exists for the example. Also can you explain how it fits overall into the GH issue summarization example? Does serving work? Did you test the trained model to see what sort of quality of results your are getting?

inc0:
Distributed training works, I still need to fix one bug in this PR to make it fully functional tho.
I was thinking of writing full fledged guide for website about keras+tfjob, because estimator approach should be 100% applicable to any tf.keras code out there, making working with tfjob much easier in general.

inc0:
>Can we replace existing code (e.g. delete some files) rather than creating yet another way to do training?

I think we could and should collapse training to single method. I just don't want to throw away t2t approach until we decide to do it together. For now, feel free to test out this approach

jlewi:
There's two ways of doing training right now

1. Using Keras
1. Using T2T

This PR is duplicating #1. So can we get rid of the existing Keras training code and just use this code?

github_issue_summarization/estimator/seq2seq_utils.py
```diff
@@ -0,0 +1,432 @@
...
+def test_gpu():
+  """Run a toy computation task in tensorflow to test GPU."""
+  config = tf.ConfigProto()
+  config.gpu_options.allow_growth = True
+  session = tf.Session(config=config)
+  hello = tf.constant('Hello, TensorFlow!')
+  print(session.run(hello))
+
+
+def plot_model_training_history(history_object):
```
jlewi: Can we not use TensorBoard?

inc0:
> This PR is duplicating #1. So can we get rid of the existing Keras training code and just use this code

What do you mean by "this code" instead of Keras? It is Keras code..

github_issue_summarization/estimator/seq2seq_utils.py
inc0: I think this is leftover from notebook, I'll clean up these from code, seq2seq utils right now is just copy paste

jlewi:
https://github.com/kubeflow/examples/blob/master/github_issue_summarization/notebooks/train.py

Is also using Keras; although I doubt its using TF.Estimator.  My question is do we need both of these? Or can we get rid of the existing notebooks/train.py?

jlewi:
SGTM. My point is I don't think we want to keep maintaining multiple versions of the code/example. So the plan should be to make this a suitable replacement so that we can get rid of what currently exists.

texasmichelle:
Yes, please replace the previous version with this one

jlewi:
@inc0 How's this coming?

texasmichelle:
I think it's time to rip out the t2t example since one has been added [here](https://github.com/kubeflow/examples/tree/master/demos/yelp_demo).

Please replace the existing code with this and include usage instructions..

jlewi:
I did another pass. It seems like there's a fair bit of code duplication with the existing Keras example.  Should that be fixed before submitting the PR?

jlewi:
I think there were some unaddressed comments on 02_distributed_training.md.

I also left some comments on tf_job.yaml; I think putting all the manifests in one file is going to be problematic.

PR #205, opened 2018-07-30
Fix T2T memory problem

### LINKED ISSUE #189: [code_search] Exploding memory with many CSV files
During the `t2t-datagen` step, exploding memory is causing the Kubeflow jobs to be killed. This issue is reproducible in local execution as well where the program terminates after running for a long time. The effect of memory pressure can also be seen in the form of decreased CPU usage - From 99% at the start of the process to around 20% towards the end before being killed.

/area example/code_search

activatedgeek:
Language problems don't really need to view all the data points but only a fraction of them. As a result, to prevent memory explosion, I've added `max_samples_for_vocab` to a reasonable limit.

- Limit the number of samples for vocab (Fixes #189)
- Update to nicer documentation
- Limit the number of CSV files to 100 to make them manageable (else all the data generates ~3k CSV files)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/205)

### DIFF
--- a/code_search/src/code_search/dataflow/cli/create_function_embeddings.py
+++ b/code_search/src/code_search/dataflow/cli/create_function_embeddings.py
@@ -17,6 +17,9 @@ def create_function_embeddings(argv=None):
     - See `transforms.github_dataset.GithubBatchPredict` for details of tables created
     - Additionally, store CSV of docstring, original functions and other metadata for
       reverse index lookup during search engine queries.
+
+  NOTE: The number of output file shards have been fixed (at 100) to avoid a large
+  number of output files, making it manageable.
   """
   pipeline_opts = arguments.prepare_pipeline_opts(argv)
   args = pipeline_opts._visible_options  # pylint: disable=protected-access
--- a/code_search/src/code_search/dataflow/cli/create_function_embeddings.py
+++ b/code_search/src/code_search/dataflow/cli/create_function_embeddings.py
@@ -37,7 +37,8 @@ def create_function_embeddings(argv=None):
     | "Format for CSV Write" >> beam.ParDo(dict_to_csv.DictToCSVString(
         ['nwo', 'path', 'function_name', 'lineno', 'original_function', 'function_embedding']))
     | "Write Embeddings to CSV" >> beam.io.WriteToText('{}/func-index'.format(args.data_dir),
-                                                       file_name_suffix='.csv')
+                                                       file_name_suffix='.csv',
+                                                       num_shards=100)
   )
 
   result = pipeline.run()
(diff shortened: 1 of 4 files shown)


code_search/src/code_search/dataflow/cli/create_function_embeddings.py
```diff
@@ -37,7 +37,8 @@ def create_function_embeddings(argv=None):
     | "Format for CSV Write" >> beam.ParDo(dict_to_csv.DictToCSVString(
         ['nwo', 'path', 'function_name', 'lineno', 'original_function', 'function_embedding']))
     | "Write Embeddings to CSV" >> beam.io.WriteToText('{}/func-index'.format(args.data_dir),
-                                                       file_name_suffix='.csv')
+                                                       file_name_suffix='.csv',
+                                                       num_shards=100)
```
jlewi: Why are you fixing the number of shards? Don't we want to let Dataflow pick the number of shards so that we don't limit parallelism?

code_search/src/code_search/t2t/function_docstring.py
```diff
@@ -1,32 +1,37 @@
...
+  def base_url(self):
+    return "gs://kubeflow-examples/t2t-code-search/raw_data"
 
-  """This class defines the problem of finding similarity between Python
-  function and docstring"""
+  @property
+  def pair_files_list(self):
+    return [
+        [
+            "{}/func-doc-pairs-000{:02}-of-00100.csv".format(self.base_url, i),
```
jlewi: Why are file paths being hard coded here?

code_search/src/code_search/dataflow/cli/create_function_embeddings.py
activatedgeek: I actually did try that but it led to ~3k files which were just too many to manage. So I fixed it to 100. I didn't find any difference in runtimes as well.

code_search/src/code_search/t2t/function_docstring.py
activatedgeek: The semantics of T2T are such that all the data and files are deterministically defined. I had earlier made them dynamic but Lukasz suggested to not do that. Instead it is easy to just derive from this class and override the property to change the file names.

code_search/src/code_search/dataflow/cli/create_function_embeddings.py
jlewi: Can you add a comment explaining this. Fixing shards is generally not considered good practice so we should note why we are doing it.

code_search/src/code_search/t2t/function_docstring.py
jlewi: Please add a comment explaining that.
Couldn't we just parameterize the derived class?

activatedgeek: Unfortunately, the T2T interface does not allow parametrization this way as it considers the dataset as an immutable. However, I will take this up with the T2T team if they have a better way of dynamically allowing this via CLI.

activatedgeek:
@jlewi I've also combined the base URL and files properties because they form one atomic unit and one can't do without the other.

PR #209, opened 2018-08-03
Merge UI and Search Server

activatedgeek:
- Use `tf.gfile` more effectively to remove manual boilerplate code.
- Use function partials to have better separation of concerns across classes
- Refactor CLI for Search Index Creation and Serving into its own `cli` module.
- Merge the UI server with the Inference API server

**NOTE**: Only the files in `code_search/src/code_search/nmslib` need to be reviewed. Everything else is just a diff from the move.

/cc @jlewi @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/209)


activatedgeek:
I will take up the final integration verification in the Kubeflow app in the next PR but should mostly be some trivial re-arrangements.

PR #213, opened 2018-08-04
Update Search Index server spec

activatedgeek:
- Update Search Index Creator Ksonnet Component
- Update Search Index Server Ksonnet Component
- The Search UI is now deployable on any url path prefix (via a `PUBLIC_URL` environment variable). This is the same environment variable used by the build process.
- Instead of building two image tags (one with the remote `gcr.io` prefix and one without), build just one based on whether `PROJECT` variable is available or not.

/cc @jlewi @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/213)

### DIFF
--- a/code_search/kubeflow/components/nms.libsonnet
+++ b/code_search/kubeflow/components/nms.libsonnet
@@ -1,9 +1,9 @@
 local baseParams = std.extVar("__ksonnet/params").components["nmslib"];
 
 {
-  spec(params, env, apiVersion="extensions/v1beta1", kind="Deployment"):: {
-    apiVersion: apiVersion,
-    kind: kind,
+  deploymentSpec(params, env, containers):: {
+    apiVersion: "extensions/v1beta1",
+    kind: "Deployment",
     metadata: {
       name: params.name,
       namespace: env.namespace,
@@ -13,7 +13,7 @@ local baseParams = std.extVar("__ksonnet/params").components["nmslib"];
     },
     spec: {
       replicas: params.replicas,
-      [if kind == "Deployment" then "selector"]: {
+      selector: {
         matchLabels: {
           app: params.name,
         },
@@ -25,31 +25,7 @@ local baseParams = std.extVar("__ksonnet/params").components["nmslib"];
           }
         },
         spec: {
-          [if kind == "Job" then "restartPolicy"]: "OnFailure",
-          containers: [
-            {
-              name: params.name,
-              image: params.image,
-              args: params.args,
-              ports: [
-                {
-                  containerPort: 8008,
-                }
-              ],
-              env: [
-                {
-                  name: "GOOGLE_APPLICATION_CREDENTIALS",
-                  value: "/secret/gcp-credentials/user-gcp-sa.json",
-                }
-              ],
-              volumeMounts: [
-                {
-                  mountPath: "/secret/gcp-credentials",
-                  name: "gcp-credentials",
-                },
-              ],
-            }
-          ],
+          containers: containers,
           volumes: [
             {
               name: "gcp-credentials",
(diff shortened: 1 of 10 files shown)


code_search/kubeflow/components/nms.libsonnet
```diff
@@ -2,7 +2,49 @@ local baseParams = std.extVar("__ksonnet/params").components["nmslib"];
 
 {
 
-  nmsContainer(params, env):: {
+  indexCreator(params, env):: {
+    apiVersion: "v1",
+    kind: "Pod",
```
jlewi: Why are you using a Pod? You should always use a controller e.g. Deployment. A Pod will not be recreated if it is terminated. But a controller like a Deployment will create a new pod if the existing pod stops running.

activatedgeek:
The service is now live at https://kf-cs.endpoints.kubeflow-dev.cloud.goog/code-search/.

code_search/kubeflow/components/nms.libsonnet
activatedgeek: Makes sense. I did try using the "Job" kind but was running into certain issues. Perhaps some spec keys from "Deployment" kind are not compatible with "Job". I'll fix this.

code_search/kubeflow/components/nms.libsonnet
```diff
@@ -2,7 +2,49 @@ local baseParams = std.extVar("__ksonnet/params").components["nmslib"];
 
 {
 
-  nmsContainer(params, env):: {
+  indexCreator(params, env):: {
```
jlewi: Do we need a libsonnet file?
The pattern we are generally following is to avoid complex ksonnet and just put the full spec in the jsonnet file so its for people to do advanced customizaiton.

code_search/kubeflow/components/nms.libsonnet
```diff
@@ -1,20 +1,19 @@
 local baseParams = std.extVar("__ksonnet/params").components["nmslib"];
 
 {
-
-  nmsContainer(params, env):: {
-    apiVersion: "extensions/v1beta1",
-    kind: "Deployment",
+  spec(params, env, apiVersion="extensions/v1beta1", kind="Deployment"):: {
```
jlewi: Why are you parameterizing the kind for this component? Why isn't nms always a Deployment or always a Job?

code_search/kubeflow/components/params.libsonnet
```diff
@@ -68,27 +68,36 @@
...
+      indexFile: $.global.t2tWorkingDir + "/code_search_index.nmslib",
     },
 
-    "nms-server": {
-      name: "nms-server",
+    "search-index-server": {
+      name: "search-index-server",
+
+      problem: "github_function_docstring",
+      dataDir: $.global.t2tWorkingDir + "/data",
```
jlewi: Do you want dataDir and lookupFile,... to always be a function of t2tWorkingDir or do you want to let users override the parameters in a way that breaks that pattern.

code_search/kubeflow/components/nms.libsonnet
activatedgeek: The step before deploying the search server is the index creation. This is a one-off job. I observed that the index was being created multiple times over. This was because when using the kind "Deployment", restartPolicy can only be "Always" (which was kind of expected in hindsight) and even if the container exited, it just restarted building the index again. "Job" allows restartPolicy of "onFailure" which made sense in this case.

code_search/kubeflow/components/params.libsonnet
activatedgeek: It is certainly possible to break this pattern but I wanted to minimize the amount of parameters the user was changing. This allows the users to keep everything under one umbrella, easier to see all artifacts in one place (one GCS directory). The advanced users might go ahead and break this pattern, and should be easy enough once they understand the mechanics.

code_search/kubeflow/components/nms.libsonnet
jlewi: I'm still confused. Lets chat about this.

PR #220, opened 2018-08-14
Add Yelp restaurant review demo files

texasmichelle:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/220)


texasmichelle:
To make it clear that it is not a DIY self-paced example like the rest of the subfolders.

jlewi:
What's the distinction between a DIY self-paced example and the other examples? Is it just the amount of documentation? 

I was imagining that information about the overall state / purpose of each example would just go in the top level README.md?
https://github.com/kubeflow/examples

/lgtm
/approve

/hold  just cancel it if you want to leave it as is.

texasmichelle:
The instructions are very different because one is designed for teaching chronologically & the other to show off specific features and nothing else, which I think warrants categorical separation. Since this was merged, I'll consider restructuring in a follow-on PR since I will have some mods to these files soon.

PR #225, opened 2018-08-14
Add a new github function docstring extended problem

### LINKED ISSUE #223: [code_search] Allow configurable T2T problem
Currently, the T2T problem is hard-coded with the dataset. This is by design and to preserve the semantics of `data_dir` (it should be the working directory into which TFRecords are written and not point to the raw data).

We instead would like to override these semantics and the easier way is to create another problem class which overrides the current `GithubFunctionDocstring` and allows the semantics of `data_dir` which suit our purpose. Ideally, the interface should involve just overriding a new class and always use that problem from the `t2t-datagen/trainer/export` command lines.

/area example/code_search

activatedgeek:
This PR provides a new problem class which appropriates the semantics of `data_dir` in T2T CLI for our use case. Fixes #223. Pointing to the data directory will now allow training models with new data (instead of the problem fetching from an immutable data source).

/cc @jlewi @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/225)


activatedgeek:
Hold on changes from #224.

/hold

PR #226, opened 2018-08-14
Cherry pick changes to PredictionDoFn

activatedgeek:
This PR takes the chunk of `PredictionDoFn` from [kubeflow/batch-predict](https://github.com/kubeflow/batch-predict) and adds ability to allow passthrough values. See https://github.com/kubeflow/batch-predict/issues/10 for reasons. 

This file should ideally be retired once the upstream design issue is resolved. This removed dependence on a private fork and keeps the Code Search example self-contained.

/cc @jlewi @ankushagarwal

### DIFF
--- a/code_search/src/code_search/dataflow/do_fns/prediction_do_fn.py
+++ b/code_search/src/code_search/dataflow/do_fns/prediction_do_fn.py
@@ -1,0 +1,215 @@
+"""A cherry-pick of PredictionDoFn from Kubeflow Batch Predict.
+
+DONOT submit revisions to this file. See kubeflow/batch-predict#10
+for design reasons. This file should ideally be retired once the upstream
+is fixed.
+"""
+
+import datetime
+import json
+import logging
+import threading
+import traceback
+
+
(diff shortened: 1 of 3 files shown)


activatedgeek:
Wait for #224 and #225 to go in first because this requires another update to the Codelab Notebook.

/hold

code_search/src/code_search/dataflow/do_fns/prediction_do_fn.py
```diff
@@ -0,0 +1,216 @@
+"""A cherry-pick of PredictionDoFn from Kubeflow Batch Predict.
+
+DONOT submit revisions to this file. See kubeflow/batch-predict#10
```
jlewi: Can you update this with a TODO that points to an issue.

activatedgeek: Do you want me to create a new issue within examples? I've pointed to the batch prediction issue https://github.com/kubeflow/batch-predict/issues/10.

PR #229, opened 2018-08-17
Upgrade notebook commands and other relevant changes

activatedgeek:
This PR comes after a complete end-to-end test of all the commands from Jupyter Notebook.

Deployed at https://dev-latest.kubeflow.org/code-search/. This deployment was done completely from the Jupyter Notebook.

- The README now recommends simply using the Terminal form Notebooks UI instead of going to the Lab UI.
- Upgrade to Tensor2Tensor `1.7.0` (had a multi-threading fix which was relevant)
- During index creation, there might be other extraneous CSV files in the folder. Read only the relevant `index` files.
- The upstream T2T package has a copy of the problem and models in this repository. To differentiate, they have been re-registered with new names.
- Reduce number of samples used by vocabulary to prevent memory explosion (was leading to `OOMKilled` pods)
- Updated Docker images
- Update all commands in Jupyter Notebook to show representative artifacts like CSV files.
- Remove non-required `CLUSTER_*` configuration variables.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/229)


activatedgeek:
@jlewi @ankushagarwal PTAL. This is ready for review.

PR #233, opened 2018-08-23
Update Similarity Transformer Architecture to compute both embeddings

activatedgeek:
This PR takes steps towards a better architecture by incorporating a conditional graph operation to allow switching between the code embedding network and the string embedding network.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/233)

### DIFF
--- a/code_search/src/code_search/t2t/query.py
+++ b/code_search/src/code_search/t2t/query.py
@@ -14,7 +14,7 @@ def get_encoder(problem_name, data_dir):
   return problem.feature_info["inputs"].encoder
 
 
-def encode_query(encoder, query_str):
+def encode_query(encoder, query_str, embed_code=False):
   """Encode the input query string using encoder. This
   might vary by problem but keeping generic as a reference.
   Note that in T2T problems, the 'targets' key is needed
@@ -22,8 +22,11 @@ def encode_query(encoder, query_str):
   See tensorflow/tensor2tensor#868"""
 
   encoded_str = encoder.encode(query_str) + [text_encoder.EOS_ID]
-  features = {"inputs": tf.train.Feature(int64_list=tf.train.Int64List(value=encoded_str)),
-              "targets": tf.train.Feature(int64_list=tf.train.Int64List(value=[0]))}
+  features = {
+      "inputs": tf.train.Feature(int64_list=tf.train.Int64List(value=encoded_str)),
+      "targets": tf.train.Feature(int64_list=tf.train.Int64List(value=[0])),
+      "embed_code": tf.constant(embed_code, dtype=tf.bool)
+  }
   example = tf.train.Example(features=tf.train.Features(feature=features))
   return base64.b64encode(example.SerializeToString()).decode('utf-8')
 
--- a/code_search/src/code_search/t2t/similarity_transformer.py
+++ b/code_search/src/code_search/t2t/similarity_transformer.py
@@ -21,46 +21,33 @@ class SimilarityTransformer(t2t_model.T2TModel):
     return body_output
 
   def body(self, features):
+    with tf.variable_scope('string_embedding'):
+      string_embedding = self.encode(features, 'inputs')
 
-    def embed_string():
-      with tf.variable_scope('string_embedding'):
-        string_embedding = self.encode(features, 'inputs')
-      return string_embedding
+    with tf.variable_scope('code_embedding'):
+      code_embedding = self.encode(features, 'targets')
 
-    def embed_code():
-      with tf.variable_scope('code_embedding'):
-        code_embedding = self.encode(features, 'targets')
-      return code_embedding
+    result = tf.concat([string_embedding, code_embedding], 1)
 
-    predicate = tf.cast(features.get('embed_code'), dtype=tf.bool)[0][0]
+    if self.hparams.mode != tf.estimator.ModeKeys.PREDICT:
+      # string_embedding_norm = tf.nn.l2_normalize(string_embedding, axis=1)
+      # code_embedding_norm = tf.nn.l2_normalize(code_embedding, axis=1)
 
-    if self.trainable:
-      string_embedding = embed_string()
-      code_embedding = embed_code()
+      p = tf.nn.sigmoid(tf.matmul(string_embedding, code_embedding,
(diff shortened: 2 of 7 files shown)


jlewi:
@activatedgeek What's the status of this PR?

Is there any way to right a unittest to make sure that this works as expected? 

What if we mock out/override encode? Would that allow us to replace a transformer encoder with something known (maybe just pass the features through?)? So we would know what the expected output is.

activatedgeek:
I've actually been using this branch to test out my variations and particularly to get a solution out for #259.

code_search/src/code_search/t2t/similarity_transformer.py
```diff
@@ -73,5 +72,9 @@ def encode(self, features, input_key):
 
   def infer(self, features=None, **kwargs):
     del kwargs
+
+    if 'targets' not in features:
```
jlewi: Why do we need to check if targets is in features? 
Is the prediction in inference mode the concatenation of code and query embeddings?
If it is then do we need this if statement because the user can control what embeddings are computed
1) setting inputs and targets to zeros or non-zero values 
2) Which elements of the output vector they look at.

code_search/src/code_search/t2t/query.py
```diff
@@ -14,16 +14,19 @@ def get_encoder(problem_name, data_dir):
   return problem.feature_info["inputs"].encoder
 
 
-def encode_query(encoder, query_str):
+def encode_query(encoder, query_str, embed_code=False):
```
jlewi: Can we get rid of embed_code now that we are concatenating the two vectors and computing both embeddings?

jlewi:
@activatedgeek It looks like this PR is making two changes

1. Change the loss function to fix #259 
1. Fix model export to allow #260 to compute embeddings

I believe #2 will necessitate changes to other parts of the code because the output of inference will now be a vector
[query_embeddings, code_embeddings]

So would it make sense to get an initial version of this PR merged so we can begin making the other changes to the code?

jlewi:
@activatedgeek And thank you so much for continuing to work on this! Great to see this coming along.

code_search/src/code_search/t2t/similarity_transformer.py
activatedgeek: The eval step during the training also uses this function and deliberately does not send in targets. I added this to make sure they work. Certainly not ideal though.

code_search/src/code_search/t2t/query.py
activatedgeek: This actually leads to a bunch of changes here and even the Dataflow step before this. Shall we defer this change to the final PR after we make sure the model works?

activatedgeek:
@jlewi I think we can go ahead and merge this.

I'm happy to be working on this, albeit slower than I'd like.

code_search/src/code_search/t2t/similarity_transformer.py
activatedgeek: It is kind of the client's responsibility to slice and dice the I/O vectors. This part is mostly safeguard the eval step of T2T.

jlewi: Can you put that in a comment?

code_search/src/code_search/t2t/query.py
jlewi: Defer which change the whole PR or just removing embed_code?

jlewi:
Regarding your earlier comment

> This actually leads to a bunch of changes here and even the Dataflow step before this. Shall we defer this change to the final PR after we make sure the model works?

Was that referring to getting rid of embed_config or this PR as a whole?

I assume once we merge this PR a lot of the Dataflow code will break and need to be updated?

Any suggestions about how we can go about fixing the code?

Would it be possible to start writing unittests to verify the various Dataflow transforms are working with the new model? Perhaps we could support that by adding a test utility function to emit dummy models for the new model architecture?

code_search/src/code_search/t2t/query.py
```diff
@@ -14,16 +14,19 @@ def get_encoder(problem_name, data_dir):
...
   even though it is ignored during inference.
   See tensorflow/tensor2tensor#868"""
 
   encoded_str = encoder.encode(query_str) + [text_encoder.EOS_ID]
-  features = {"inputs": tf.train.Feature(int64_list=tf.train.Int64List(value=encoded_str)),
-              "targets": tf.train.Feature(int64_list=tf.train.Int64List(value=[0]))}
+  features = {
+      "inputs": tf.train.Feature(int64_list=tf.train.Int64List(value=encoded_str)),
+      "targets": tf.train.Feature(int64_list=tf.train.Int64List(value=[0])),
+      "embed_code": tf.constant(embed_code, dtype=tf.bool)
```
jlewi: This looks like a bug. I think FeatureProto can only be Int,Bytes, and Floats.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/example/feature.proto

Should "inputs" be the concatenation of the encoded query and the encoded targets now that we are treating output in prediction as concatenation of the two vectors?

code_search/src/code_search/t2t/similarity_transformer.py
jlewi: I think this is causing problems during inference using TFServing.

When I send predictions to TFServing I'm getting b'{ "error": "You must feed a value for placeholder tensor \\\'Placeholder\\\' with dtype int64 and shape [?,?,1,1]\\n\\t [[{{node Placeholder}} = Placeholder[_output_shapes=[[?,?,1,1]], dtype=DT_INT64, shape=[?,?,1,1], _device=\\"/job:localhost/replica:0/task:0/device:CPU:0\\"]()]]" }'

jlewi:
I think we can close this in favor of #291.

I originally based #291 on this PR but I think I fixed the issues with the using conditionals.

1. The inputs will always be provided by the "inputs" feature; we will just use different embedding variable scope
1. We need to reduce the feature used as the predicate to a rank 0 vector. I think this may have caused the problems during eval.

activatedgeek:
I think this PR has diverged enough to be closed and all the newer code is more relevant anyways. Let me close this.

PR #242, opened 2018-09-03
Upgrade demo to use latest versions of kubeflow, tfjob, ksonnet, & gke

texasmichelle:
Upgrade kubeflow to v0.2.4
Upgrade ksonnet app dir to v0.12.0
Update t2t components to use tfjob-v1alpha2
Upgrade gke cluster version to 1.10.6-gke.2 on creation
Add instructions for creating a single cluster with GPU & TPU

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/242)


texasmichelle:
Does it make sense to remove lib files for older k8s versions in demos/yelp_demo/demo/lib/ksonnet-lib? Should we add these to .gitignore? @jlewi

jlewi:
What's the point of the v1alpha1 files? Why not just delete the v1alpha1 versions at this point?

Just an FYI about ksonnet 0.12; apps in ksonnet 0.12 can't be used with ksonnet 0.11; however ksonnet 0.12 will work with 0.11 apps. 

I don't think you want to add lib to .gitignore; that might break the app when its checked out and those files are missing. (Although I suspect those files might get added when the user adds an environment corresponding to their K8s cluster).

jlewi:
@texasmichelle left a couple comments. You can consider them optional; just cancel the hold.
/lgtm
/approve
/hold 
/assign @jlewi
If I don't approve by tonight you should reassign to someone else

texasmichelle:
The service account creation is for non-GKE environments. In those cases, t2tcpu uses GOOGLE_APPLICATION_CREDENTIALS and also needs a way to pull images from private gcr.io registries. The VM service account approach only works on GKE. Since TPUs on GKE are still in beta, I couldn’t figure out how to create a cluster with Deployment Manager and make use of the "{DEPLOYMENT_NAME}-user" account.

I deleted all of the *-v1alpha1.jsonnet files, so maybe that was an outdated reference. They should all be gone now.

I created #244 for replacing the cluster.jinja file, #243 for using the deploy script, and #245 for refactoring t2t jsonnet files.

PR #246, opened 2018-09-05
example mnist upgrade to v1alpha2

fisache:
upgrade tf-operator from v1alpha1 to v1alpha2 in mnist example

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/246)

### DIFF
--- a/mnist/model-train.yaml
+++ b/mnist/model-train.yaml
@@ -120,7 +120,6 @@ spec:
           name: {{workflow.parameters.job-name}}
           namespace: {{workflow.parameters.namespace}}
         spec:
-          cleanPodPolicy: All
           tfReplicaSpecs:
             Master:
               replicas: 1
@@ -210,7 +209,6 @@ spec:
                   restartPolicy: OnFailure
             Ps:
               replicas: {{workflow.parameters.tf-ps}}
-              tfReplicaType: "PS"
               template:
                 spec:
                   containers:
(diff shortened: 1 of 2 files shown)


mnist/model-train.yaml
```diff
@@ -112,147 +112,145 @@ spec:
...
-                    - name: S3_VERIFY_SSL
-                      value: "{{workflow.parameters.s3-verify-ssl}}"
-                    - name: S3_ENDPOINT
-                      value: {{workflow.parameters.s3-endpoint}}
-                restartPolicy: OnFailure
-          terminationPolicy:
-            chief:
-              replicaIndex: 0
-              replicaName: MASTER
+          cleanPodPolicy: All
```
jlewi: I would recommend not setting cleanPodPolicy; this will use the default which only delete running pods so completed pods are left alone to preserve the logs.

jlewi:
Thanks for the fix.
One minor change requested.

/assign @texasmichelle 

@texasmichelle would mind approving this after the requested changes are made since I'll be OOO.

mnist/model-train.yaml
fisache: @jlewi Thx, I removed it

fisache:
when use master branch, I get error installing argo.
then, I fix branch to v0.2

texasmichelle:
Thanks for updating this example!

/lgtm
/approve

PR #255, opened 2018-09-29
Add tensorboard and check in vendor for the code search example.

jlewi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/255)

### DIFF
--- a/code_search/kubeflow/components/params.libsonnet
+++ b/code_search/kubeflow/components/params.libsonnet
@@ -81,8 +81,8 @@
     },
     tensorboard: {
       image: "tensorflow/tensorflow:1.8.0",
-      logDir: "gs://cloud-ml-dev_jlewi_cs/mode/2018_0825/",
+      logDir: "gs://example/to/model/logdir",
       name: "tensorboard",
-    }, 
+    },
   },
 }
--- a/code_search/kubeflow/components/params.libsonnet
+++ b/code_search/kubeflow/components/params.libsonnet
@@ -79,5 +79,10 @@
       indexFile: $.components['t2t-code-search'].workingDir + '/code_search_index.nmslib',
       servingUrl: 'http://t2t-code-search.kubeflow:9001/v1/models/t2t-code-search:predict',
     },
+    tensorboard: {
+      image: "tensorflow/tensorflow:1.8.0",
+      logDir: "gs://cloud-ml-dev_jlewi_cs/mode/2018_0825/",
+      name: "tensorboard",
+    }, 
   },
 }
--- a/code_search/kubeflow/components/t2t-job.libsonnet
+++ b/code_search/kubeflow/components/t2t-job.libsonnet
@@ -21,35 +21,35 @@ local baseParams = std.extVar("__ksonnet/params").components["t2t-job"];
     ],
 
   getTrainerCmd(params):: {
-      local trainer = [
-        "/usr/local/sbin/t2t-entrypoint",
-        "t2t-trainer",
-        "--problem=" + params.problem,
-        "--model=" + params.model,
-        "--hparams_set=" + params.hparams_set,
-        "--data_dir=" + params.dataDir,
-        "--output_dir=" + params.outputDir,
-        "--train_steps=" + std.toString(params.train_steps),
-        "--eval_steps=" + std.toString(params.eval_steps),
-        "--t2t_usr_dir=/app/code_search/t2t",
-      ],
+    local trainer = [
+      "/usr/local/sbin/t2t-entrypoint",
+      "t2t-trainer",
+      "--problem=" + params.problem,
+      "--model=" + params.model,
+      "--hparams_set=" + params.hparams_set,
+      "--data_dir=" + params.dataDir,
+      "--output_dir=" + params.outputDir,
+      "--train_steps=" + std.toString(params.train_steps),
+      "--eval_steps=" + std.toString(params.eval_steps),
+      "--t2t_usr_dir=/app/code_search/t2t",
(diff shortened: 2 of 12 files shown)


activatedgeek:
/lgtm

It appears that most changes to the Ksonnet files are just reformatting

code_search/kubeflow/vendor/kubeflow/examples@defc235463799d5600001ee0ed6ef68f7af24a17/prototypes/tf-batch-predict.jsonnet
```diff
@@ -0,0 +1,143 @@
+// @apiVersion 0.1
```
inc0: should this be in different PR?

code_search/kubeflow/vendor/kubeflow/examples@defc235463799d5600001ee0ed6ef68f7af24a17/prototypes/tf-job-simple-v1alpha1.jsonnet
```diff
@@ -0,0 +1,91 @@
+// @apiVersion 0.1
```
inc0: ditto, seems like component from different problem

code_search/kubeflow/vendor/kubeflow/examples@defc235463799d5600001ee0ed6ef68f7af24a17/prototypes/tf-serving-with-istio.jsonnet
```diff
@@ -0,0 +1,179 @@
+// @apiVersion 0.1
```
inc0: If you want to add tf-serving + istion component, let's make dedicated PR

code_search/kubeflow/components/params.libsonnet
```diff
@@ -79,5 +79,10 @@
       indexFile: $.components['t2t-code-search'].workingDir + '/code_search_index.nmslib',
       servingUrl: 'http://t2t-code-search.kubeflow:9001/v1/models/t2t-code-search:predict',
     },
+    tensorboard: {
+      image: "tensorflow/tensorflow:1.8.0",
+      logDir: "gs://cloud-ml-dev_jlewi_cs/mode/2018_0825/",
```
inc0: do you want to keep your personal path here?

code_search/kubeflow/components/t2t-job.libsonnet
```diff
@@ -84,7 +84,7 @@ local baseParams = std.extVar("__ksonnet/params").components["t2t-job"];
     local workerEnv = [
       {
         name: "GOOGLE_APPLICATION_CREDENTIALS",
-        value: "/secret/gcp-credentials/user-gcp-sa.json"
+        value: "/secret/gcp-credentials/user-gcp-sa.json",
```
inc0: can you add another env variable where you point to log dir? Then we could refer to it in code and allow everyone to use tensorboard with this example

code_search/kubeflow/vendor/kubeflow/examples@defc235463799d5600001ee0ed6ef68f7af24a17/prototypes/tf-batch-predict.jsonnet
jlewi: I'm not following? You want one PR for adding vendor and one for adding TB?

jlewi: The TensorBoard prototoype comes from the examples package. So we need to install the entire examples package in order to get TB which brings in other prototypes as well.

code_search/kubeflow/components/t2t-job.libsonnet
jlewi: Can you explain what you are asking? Everyone can use tensorboard with this example. They just configure their TB job to point at the same directory that their T2T job writes the model.

jlewi:
Thanks PTAL

code_search/kubeflow/components/t2t-job.libsonnet
inc0: I was thinking of allowsing something like `tf.summary.FileWriter(os.getenv("KUBEFLOW_LOGDIR"), sess.graph))`

code_search/kubeflow/vendor/kubeflow/examples@defc235463799d5600001ee0ed6ef68f7af24a17/prototypes/tf-batch-predict.jsonnet
inc0: ok, makes sense, didn't notice we add whole component

code_search/kubeflow/components/t2t-job.libsonnet
jlewi: Why do we need that? The trainer code takes command line argument modelDir which controls where the model and event files are written. So if the user wants to change the location of the event files. They can just change that.

jlewi:
@inc0 Ping?

code_search/kubeflow/components/t2t-job.libsonnet
inc0: ahh that makses sense. I've missed that.

jlewi:
Thanks
/approve

PR #270, opened 2018-10-17
Fix bash

Oblynx:
Remove spaces around a bash variable declaration.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/270)


Oblynx:
I signed it!

PR #278, opened 2018-10-22
Upgrade demo to KF v0.3.1

texasmichelle:
Update env variable names and values in base file
Cleanup ambassador metadata for UI component
Add kfctl installation instructions
Tighten minikube setup instructions and update k8s version
Move environment variable setup to very beginning
Replace cluster creation commands with links to the appropriate section in demo_setup/README.md
Replace deploy.sh with kfctl
Replace kubeflow-core component with individual components
Remove connection to UI pod directly & connect via ambassador instead
Add cleanup commands

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/278)


demos/yelp_demo/README.md
```diff
@@ -119,13 +95,13 @@ kubectl config use gke
...
 ```
-ks env add gke
+ks env add gke --namespace=kubeflow
 ```
 
 Install kubeflow on the cluster:
 
 ```
-ks apply gke -c kubeflow-core
+ks apply gke -c ambassador -c centraldashboard -c jupyterhub -c spartakus -c tf-job-operator
```
lluunn: why do separate components instead of core?

demos/yelp_demo/README.md
```diff
@@ -209,43 +185,34 @@ Create the serving and UI components:
...
 ```
-UI_POD=$(kubectl get po -l app=kubeflow-demo-ui | \
-  grep kubeflow-demo-ui | \
+AMBASSADOR_POD=$(kubectl get po -l service=ambassador | \
+  grep ambassador | \
   head -n 1 | \
   cut -d " " -f 1 \
 )
-kubectl port-forward ${UI_POD} 8080:80
+kubectl port-forward ${AMBASSADOR_POD} 8080:80
```
lluunn: Why port-foward ambassador? Are we using it like ingress?

demos/yelp_demo/README.md
texasmichelle: `kfctl` creates them as individual components, so running `ks apply gke -c kubeflow-core` is a no-op

demos/yelp_demo/README.md
texasmichelle: Isn't that the point of the reverse proxy - to port-forward once and access various components via pathname?

lluunn: I guess for the demo purpose use ambassador as a reverse proxy should be fine.

Issue #258, opened 2018-09-30
[code_search] Unittest for similarity and loss computation using TF.Eager 


jlewi:
Here:
https://github.com/kubeflow/examples/blob/2064b43deffa436fdf148edc2d4e67ee1580839a/code_search/src/code_search/t2t/similarity_transformer.py#L32

We construct a simple network to compute the loss for the query and code embeddings.

We can pull that code out into a class method and then write a unittest for it using TF.eager

cwbeitel:
Here's a starting point:

PR #291, opened 2018-11-01
Use conditionals and add test for code search

### LINKED ISSUE #260: [code_search] Fix model export for computing code embeddings
Right now I think we are using the same T2T exported model for computing the code embeddings and the function embeddings.

But I think T2T exporter will create an exported model that outputs whatever the body function returns as outputs (see [T2T docs](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/utils/t2t_model.py#L67))

Our body function is returning the query/docstring embeddings 
https://github.com/kubeflow/examples/blob/master/code_search/src/code_search/t2t/similarity_transformer.py#L52

Does #233 partially address this by adding a conditional variable to control which embeddings to return?

/cc @activatedgeek @cwbeitel

jlewi:
Fix model export, loss function, and add some manual tests.

Fix Model export to support computing code embeddings: Fix #260

* The previous exported model was always using the embeddings trained for
  the search query.

* But we need to be able to compute embedding vectors for both the query
  and code.

* To support this we add a new input feature "embed_code" and conditional
  ops. The exported model uses the value of the embed_code feature to determine
  whether to treat the inputs as a query string or code and computes
  the embeddings appropriately.

* Originally based on #233 by @activatedgeek

Add manual tests

* Related to #258

* We add a smoke test for T2T steps so we can catch bugs in the code.
* We also add a smoke test for serving the model with TFServing.
* We add a sanity check to ensure we get different values for the same
  input based on which embeddings we are computing.

Change Problem/Model name

* Register the problem github_function_docstring with a different name
  to distinguish it from the version inside the Tensor2Tensor library.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/291)


jlewi:
/hold cancel
/assign @cwbeitel @activatedgeek  

This is ready for review.

cwbeitel:
@jlewi Various minor comments, none blocking, ready to approve once the presubmit tests pass, looks like there are some minor linter errors and it can't run the tests b/c tensorflow can't be imported.

cwbeitel:
If the tests passed locally it's probably fine, fixing a CI dependency is kind of a separate issue.

/lgtm

cwbeitel:
Nice.

/lgtm

cwbeitel:
@jlewi Oh and just noticed the normalization at the end of the inference step.  I don't know if that's strictly necessary given nmslib is [later](https://github.com/kubeflow/examples/blob/master/code_search/src/code_search/nmslib/search_engine.py#L38) finding nearest-neighbors using the cosine distance. Which according to their manual is the standard form i.e. equivalent to tf.matmul(tf.nn.l2_normalize(a, axis=1), tf.nn.l2_normalize(b, axis=1), transpose_b=True).

jlewi:
@cwbeitel agree its not strictly necessary; but since our training metric using normalized vectors it seemed better to also use normalized vectors in inference. Otherwise it seems like users could easily forget to do that at inference time (e.g. if not using nmslib) and that could cause bugs.

Issue #308, opened 2018-11-05
[code_search] jsonnet overrides don't work as expected


jlewi:
The ksonnet is structured such that we define two dictionaries of common parameters "t2t-job"
and "t2t-code-search"
https://github.com/kubeflow/examples/blob/master/code_search/kubeflow/components/params.libsonnet#L4

These parameters can then be used by a variety of different components.

This works as long as we use the default parameters or change the parameter for all environments.

What doesn't seem to work is overriding these parameters on an environment specific basis.

For example, we'd like to be able to override t2t-job["outputDir"] for a specific environment "myenv" and then have all components use that value in environment "myenv"; but that doesn't appear to happen.

jlewi:
I wonder if the correct thing to do here is to take advantage of global parameters.
https://github.com/ksonnet/ksonnet/blob/master/docs/concepts.md

Right now we're defining pseudo modules. Here's the relevant jsonnet code in params.libsonnet for my environment

```
local envParams = params + {
  components+: {
    "t2t-code-search"+: {
      workingDir: 'gs://code-search-demo/20181103',
    },
    "t2t-code-search-datagen"+: {
      workingDir: 'gs://code-search-demo/20181104',
      dataDir: 'gs://code-search-demo/20181104/data',
      project: 'code-search-demo',
      githubTable: '',
    },
    "submit-preprocess-job"+: {
      workingDir: 'gs://code-search-demo/20181104',
      dataDir: 'gs://code-search-demo/20181104/data',
      project: 'code-search-demo',
      githubTable: '',
    },
  },
};

{
  components: {
    [x]: envParams.components[x] + globals
    for x in std.objectFields(envParams.components)
  },
}
```

So if we set a global I think it will automatically be added to every component and I think we can specify globals per environment.

This seems much easier to reason about then what we are currently doing which is relying on late binding.

jlewi:
Downside of using globals is that they can't be easily set via command line (at least not with 0.12).

kkasravi:
> The ksonnet is structured such that we define two dictionaries of common parameters "t2t-job"
> and "t2t-code-search"
> https://github.com/kubeflow/examples/blob/master/code_search/kubeflow/components/params.libsonnet#L4
> 
> These parameters can then be used by a variety of different components.
> 
> This works as long as we use the default parameters or change the parameter for all environments.
> 
> What doesn't seem to work is overriding these parameters on an environment specific basis.
> 
[...]

PR #302, opened 2018-11-05
Fix performance of dataflow preprocessing job.

jlewi:
* Fix #300; Dataflow job for preprocessing is really slow.

  * The problem is we are loading the spacy tokenization model on every
    invocation of the tokenization function and this is really expensive.
  * We should be doing this once per module import.

* After fixing this issue; the job completed in approximately 20 minutes using
  5 workers.

  * We can process all 1.3 million records in ~ 20 minutes (elapsed time) using 5 32 CPU workers and about 1 hour of CPU time altogether.

* Add options to the Dataflow job to read from files as opposed to BigQuery
  and to skip BigQuery writes. This is useful for testing.

* Add a "unittest" that verifies the Dataflow preprocessing job can run
  successfully using the DirectRunner.

* Update the Docker image and a ksonnet component for a K8s job that
  can be used to submit the Dataflow job.

* Fix #299; Add logging to the Dataflow preprocessing job to indicate that
  a Dataflow job was submitted.

* Add an option to the preprocessing Dataflow job to read an entire
  BigQuery table as the input rather than running a query to get the input.
  This is useful in the case where the user wants to run a different
  query to select the repo paths and contents to process and write them
  to some table to be processed by the Dataflow job.

/assign @activatedgeek @richardsliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/302)

PR #309, opened 2018-11-06
Update ksonnet for datagen

jlewi:
* We should use a K8s job rather than a TFJob. We can also simplify the
  ksonnet by just putting the spec into the jsonnet file rather than trying
  to share various bits of the spec with the TFJob for training.

Related to #308 use globals to allow parameters to be shared
across components (e.g. working directory.)

* Update the README with information about data.


jlewi:
/hold
#302 should be submitted first.

jlewi:
#302 is submitted.

/hold cancel
/assign @activatedgeek

code_search/demo/README.md
```diff
@@ -3,4 +3,32 @@
...
+* **project**: code-search-demo
+* **[code-search-team@kubeflow.org](https://github.com/kubeflow/internal-acls/blob/master/code-search-team.members.txt)** Google group administering access
+
+# Results
+
+## 2018-11-05
+
+jlewi@ ran experiments that produced the following results
+
+| What | location | Description
```
richardsliu: Table doesn't render correctly. I think you also need a vertical bar on the right side.

PR #274, opened 2018-10-17
[WIP] E2E Pytorch mnist example

dsdinter:
Add new example E2E Pytorch MNIST, presented at Oreilly AI London
- [ ] Create postsubmit workflow in Argo to push images to gcr.io/kubeflow-examples
- [ ] Create Markdown files with instructions

PR #316, opened 2018-11-07
Make the mnist example work with latest kubeflow

chenzhiwei:
A better look of the changes: https://github.com/chenzhiwei/kubeflow-examples/tree/mnist-model/mnist

1. Update the ks version to fix the un-standard k8s version issue https://github.com/kubeflow/kubeflow/issues/521
2. Update the tensorflow version to v1.11.0
3. Add minio as s3 alternative storage backend
4. Pre-build kubeflow tf-serving code to ksonnect image(frequently fetch from github will cause github api limit issue)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/316)


jlewi:
Thanks for doing this please? Can you update the PR description to provide some information about the changes needed? This will make it easier to review the PR.

mnist/README.md
```diff
@@ -1,170 +1,99 @@
-# Training MNIST using Kubeflow, S3, and Argo.
+# Training MNIST using Kubeflow, S3(Minio), and Argo.
```
jlewi: Why is S3 qualified using Minio? Does it not use S3 directly?

jlewi:
@elsonrodriguez Since you created the initial PR do you want to review this?

mnist/README.md
```diff
@@ -1,170 +1,99 @@
...
 
 ## Prerequisites
 
 Before we get started there a few requirements.
 
-### Kubernetes Cluster Environment
+## 1.Kubernetes Cluster Environment
 
-Your cluster must:
+* Kubernetes v1.9.1+
```
jlewi: Can we just link to the Kubeflow getting started guide rather than duplicating the instructions here?

mnist/README.md
```diff
@@ -1,170 +1,99 @@
...
-## Preparing your Kubernetes Cluster
-
-With our data and workloads ready, now the cluster must be prepared. We will be deploying the TF Operator, and Argo to help manage our training job.
-
-In the following instructions we will install our required components to a single namespace.  For these instructions we will assume the chosen namespace is `tfworkflow`:
-
-### Deploying Tensorflow Operator and Argo.
 
-We are using the Tensorflow operator to automate the deployment of our distributed model training, and Argo to create the overall training pipeline. The easiest way to install these components on your Kubernetes cluster is by using Kubeflow's ksonnet prototypes.
+## 5.Download argo
```
jlewi: Do you mean the Argo CLI?

mnist/README.md
```diff
@@ -1,170 +1,99 @@
...
-cd -
-
-# Create a user for our workflow
-kubectl apply -f tf-user.yaml
+curl -sSL -o /usr/local/bin/argo https://github.com/argoproj/argo/releases/download/v2.2.1/argo-linux-amd64
+chmod +x /usr/local/bin/argo
 ```
 
-You can check to make sure the components have deployed:
+## 6.Create tf service account
```
jlewi: What is this service account for?

mnist/README.md
```diff
@@ -181,23 +110,14 @@ The argo workflow can be daunting, but basically our steps above extrapolate as
...
-First we need to set a few variables in our workflow. Make sure to set your docker registry or remove the `IMAGE` parameters in order to use our defaults:
-
 ```
-DOCKER_BASE_URL=docker.io/elsonrodriguez # Put your docker registry here
 export S3_DATA_URL=s3://${BUCKET_NAME}/data/mnist/
 export S3_TRAIN_BASE_URL=s3://${BUCKET_NAME}/models
-export JOB_NAME=myjob-$(uuidgen  | cut -c -5 | tr '[:upper:]' '[:lower:]')
-export TF_MODEL_IMAGE=${DOCKER_BASE_URL}/mytfmodel:1.7
+export JOB_NAME=myjob1
+export TF_MODEL_IMAGE=siji/mnist-model:v1.11.0
```
jlewi: How and where is this model container big built? Can we host it in kubeflow-examples?
You can push images to that repo from our CI system
https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#image-auto-release

mnist/README.md
```diff
@@ -216,51 +136,56 @@ argo submit model-train.yaml -n ${NAMESPACE} --serviceaccount tf-user \
...
-### Tensorboard
+## 10.Using Tensorflow serving
 
-Tensorboard is deployed just before training starts. To connect:
+### Install client requirements
 
 ```
-PODNAME=$(kubectl get pod -l app=tensorboard-${JOB_NAME} -o jsonpath='{.items[0].metadata.name}')
-kubectl port-forward ${PODNAME} 6006:6006
+apt install python-pip python-setuptools --no-install-recommends
```
jlewi: Why do you need to install things on the client?

mnist/README.md
```diff
@@ -216,51 +136,56 @@ argo submit model-train.yaml -n ${NAMESPACE} --serviceaccount tf-user \
...
+pip install -r requirements.txt
 ```
 
-Tensorboard can now be accessed at [http://127.0.0.1:6006](http://127.0.0.1:6006).
+### Web client: Mnist Digit Reader
+
+```
+cd mnist-webapp
+SERVICE_IP=$(kubectl -n ${NAMESPACE} get service -l app=mnist-${JOB_NAME} -o jsonpath='{.items[0].spec.clusterIP}')
```
jlewi: Ho are you connecting to the service? Did you mean to get rid of the port-forward command?

mnist/dockerfile/Dockerfile.ksonnet
```diff
@@ -0,0 +1,18 @@
+# This container is for running ksonnet within Kubernetes
```
jlewi: Where in the example does running ksonnet in cluster come into play?

mnist/dockerfile/README.md
```diff
@@ -0,0 +1,17 @@
+# Dockerfiles
+
+These two Dockerfiles are used to build ksonnect and mnist model docker images.
+
+
+## ksonnect
+
+```
+docker build -t siji/ksonnet:v0.13.0 -f Dockerfile.ksonnet .
```
jlewi: Can we push them to kubeflow-examples by adding them to our CI system
https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#image-auto-release

mnist/mnist-webapp/app.py
```diff
@@ -0,0 +1,58 @@
...
+        img_size = 28, 28
+        image_url = request.values['imageBase64']
+        image_string = re.search(r'base64,(.*)', image_url).group(1)
+        image_bytes = io.BytesIO(base64.b64decode(image_string))
+        image = Image.open(image_bytes)
+        image = image.resize(img_size, Image.LANCZOS)
+        image = image.convert('1')
+        int_image = np.array(image)
+        image = np.reshape(int_image, 784).astype(np.float32)
+        # image = image.convert('1')
```
jlewi: Delete commented out code.

mnist/mnist-webapp/app.py
```diff
@@ -0,0 +1,58 @@
+# taken from https://community.canvaslms.com/thread/2595
```
jlewi: Can we add a unittest for this?

mnist/README.md
chenzhiwei: in this example, I changed the s3 to minio since Minio is more easier for beginners and it does not depends on AWS account anymore.

mnist/README.md
chenzhiwei: yes, linked to kubeflow.

mnist/README.md
chenzhiwei: yes, changed to argo cli.

mnist/README.md
chenzhiwei: added more description about this serviceaccount.

mnist/README.md
chenzhiwei: I prefer to change this one by one since I don't want make PR depends on each other. after this PR get merged, I can submit new PR to fix the image issue.

mnist/README.md
chenzhiwei: because the cli and webapp depend on them.

mnist/README.md
chenzhiwei: yes, get rid of the port-forward command, all things are done inside the server.

It is a bit more difficult for beginners to configure and run kubectl on their desktop/laptop.

mnist/dockerfile/Dockerfile.ksonnet
chenzhiwei: in the model-train.yaml.

mnist/dockerfile/README.md
chenzhiwei: I'd like to change this in follow up PRs.

mnist/mnist-webapp/app.py
chenzhiwei: yes, deleted.

mnist/mnist-webapp/app.py
chenzhiwei: don't think so.

mnist/dockerfile/README.md
jlewi: Can you create an issue for this please?

jlewi:
Why add MinIO to the example? If the goal is to provide an alternative to S3/GCS why not use a Kubernetes PV or PVC?

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
+## 2.ksonnect cli
 
 ```
-# Create a resource group
-az group create -n kubeflowrg -l eastus
-# Deploy the cluster
-az aks create -n kubeflowaks -g kubeflowrg -l eastus -k 1.9.6 -c 3 -s Standard_NC6
-# Authentication into the cluster
-az aks get-credentials -n kubeflowaks -g kubeflowrg
+curl -fksSL https://github.com/ksonnet/ksonnet/releases/download/v0.13.0/ks_0.13.0_linux_amd64.tar.gz \
```
jlewi: Isn't downloading and installing ksonnet part of the Kubeflow getting started guide? Do we need to repeat it here?

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
 
-NOTE: You must be a Kubernetes admin to follow this guide. If you are not an admin, please contact your local cluster administrator for a client cert, or credentials to pass into the following commands:
+## 3.Kubeflow environment
 
 ```
-$ kubectl config set-credentials <username> --username=<admin_username> --password=<admin_password>
-$ kubectl config set-context <context_name> --cluster=<cluster_name> --user=<username> --namespace=<namespace>
-$ kubectl config use-context <context_name>
+git clone -b v0.3.2 https://github.com/kubeflow/kubeflow
+./kubeflow/scripts/kfctl.sh init kf-app --platform none
```
jlewi: Why repeat these instructions as opposed to just referring to the getting started guide?

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
-- [argo](https://github.com/argoproj/argo/blob/master/demo.md#1-download-argo)
-- [ksonnet](https://ksonnet.io/#get-started)
 
-To run the client at the end of the example, you must have [requirements.txt](requirements.txt) intalled in your active python environment.
+## 4.Setup Minio
 
-```
-pip install -r requirements.txt
-```
+Minio provides an S3 compatible API, we will use Minio in this example to act as S3.
```
jlewi: Per other comment; if we want an alternative to S3 why not just use a PV?

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
 
-NOTE: These instructions rely on Github, and may cause issues if behind a firewall with many Github users. Make sure you [generate and export this token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/):
+### Deploy Minio
 
 ```
-export GITHUB_TOKEN=xxxxxxxx
+docker run -e MINIO_ACCESS_KEY=minio -e MINIO_SECRET_KEY=minio123 \
```
jlewi: Are you running minio locally? Shouldn't you be deploying it on the cluster? How are pods running in the cluster supposed to access a locally running docker container?

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
+cd kf-app
+../kubeflow/scripts/kfctl.sh generate k8s
+../kubeflow/scripts/kfctl.sh apply k8s
 ```
 
-### Local Setup
+For detail info, check here: https://www.kubeflow.org/docs/started/getting-started/
 
-You also need the following command line tools:
+Note: The vizier-db in Kubeflow depends on a PV, you need to create it in order to make vizier-db work. But this does not affect this example.
```
jlewi: Are you saying that when you deploy vizier it doesn't deploy out of the box automatically? Is there an issue for this?

mnist/README.md
jlewi: Did you forget to add tf-user.yaml to the PR?

mnist/README.md
jlewi: Can you open an issue please to track this and reference it in a TODO?

mnist/README.md
```diff
@@ -216,51 +138,55 @@ argo submit model-train.yaml -n ${NAMESPACE} --serviceaccount tf-user \
...
+$ argo -n ${NAMESPACE} get tf-workflow-h7hwh
 ```
 
-## Monitoring
+After the STATUS to `Succeeded`, then you can use it.
 
-There are various ways to monitor workflow/training job. In addition to using `kubectl` to query for the status of `pods`, some basic dashboards are also available.
 
-### Argo UI
+## 9.Submit serving workflow[optional]
```
jlewi: What is the serving workflow? And why does model serving need to be disabled?

mnist/README.md
jlewi: Why do they need to cd to the directory mnist-webapp?
How can you run the port-forward command inside the server?

mnist/mnist-webapp/README.md
```diff
@@ -0,0 +1,23 @@
+# Mnist Digit Reader
```
jlewi: Why do we need to copy the mnist-webapp code to the example repo? Is there a docker container already that we can use?

mnist/model-train.yaml
```diff
@@ -15,13 +15,13 @@ spec:
     - name: tf-ps # number of tf parameter servers
       value: 2
     - name: tf-model-image
-      value: elsonrodriguez/mytfmodel:1.7
-    - name: tf-serving-image #FIXME this image is a mirror of a private kubeflow-ci image, once we're building images swap this out. https://github.com/kubeflow/kubeflow/blob/dcf4adfe2dd1cec243647f3dd05d7c26246fddb1/components/k8s-model-server/images/Dockerfile.cpu
-      value: elsonrodriguez/model-server:1.6
+      value: siji/mnist-model:v1.11.0
```
jlewi: What is Argo being used for in this example?

mnist/dockerfile/ksonnet-entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+#FIXME this is set to tf-user by default in order to work around lack of Serviceaccount setting in argo.
+SERVICE_ACCOUNT=${SERVICE_ACCOUNT:-tf-user}
```
jlewi: Do we still need this file it looks like you got rid of the container?

mnist/dockerfile/Dockerfile.ksonnet
jlewi: So this is so that we can have an Argo workflow that creates a ksonnet app to deploy the model?

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
 
 Before we get started there a few requirements.
 
-### Kubernetes Cluster Environment
+## 1.Kubernetes Cluster Environment
```
elsonrodriguez: The numerical formatting makes the document difficult to read. It now looks like every step in the guide is a prerequisite.

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
-- Contain 3 nodes of at least 8 cores and 16 GB of RAM.
-
-If using GKE, the following will provision a cluster with the required features:
-
-```
-export CLOUDSDK_CONTAINER_USE_CLIENT_CERTIFICATE=True
-gcloud alpha container clusters create ${USER} --enable-kubernetes-alpha --machine-type=n1-standard-8 --num-nodes=3 --disk-size=200 --zone=us-west1-a --cluster-version=1.9.3-gke.0 --image-type=UBUNTU
-```
-
-If using Azure, the following will provision a cluster with the required features, [using the az cli](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest):
```
elsonrodriguez: This section is compact enough that I feel there's some justification to keep it.  Even though it repeats in essence some of the info in the official linked guide, it's more to the point for the purposes of the example, and it also includes Azure instructions, which are currently missing from the official getting started guide.

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
 
-You also need the following command line tools:
+Note: The vizier-db in Kubeflow depends on a PV, you need to create it in order to make vizier-db work. But this does not affect this example.
 
-- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
-- [argo](https://github.com/argoproj/argo/blob/master/demo.md#1-download-argo)
-- [ksonnet](https://ksonnet.io/#get-started)
 
-To run the client at the end of the example, you must have [requirements.txt](requirements.txt) intalled in your active python environment.
+## 4.Setup Minio
```
elsonrodriguez: I feel that recommending minio (while it is an excellent option) as the default setup is not setting the right tone for pragmatic engineering. Unless there are specific requirements, or your environment has no S3-compatible object store, there isn't a need for Minio. It just becomes one more moving piece in an already complicated example.

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
-NOTE: These instructions rely on Github, and may cause issues if behind a firewall with many Github users. Make sure you [generate and export this token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/):
+### Deploy Minio
 
 ```
-export GITHUB_TOKEN=xxxxxxxx
+docker run -e MINIO_ACCESS_KEY=minio -e MINIO_SECRET_KEY=minio123 \
+    --name=minio -d --net=host -v /var/lib/minio:/data minio/minio server /data
 ```
 
-## Modifying existing examples
```
elsonrodriguez: It actually took a bit of work to get the upstream mnist example to work decently. There's educational value in highlighting some of the shortcomings of the canned examples floating around in regards to configurability and scalability.

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
 
-The resulting model is [model.py](model.py).
-
-### Build and push model image.
-
-With our code ready, we will now build/push the docker image.
+### Create Minio bucket
 
 ```
-DOCKER_BASE_URL=docker.io/elsonrodriguez # Put your docker registry here
```
elsonrodriguez: Removing this step makes the origin of the model container opaque, as stated by Jeremy in one of his reviews.

mnist/README.md
```diff
@@ -6,165 +6,96 @@ This example guides you through the process of taking an example model, modifyin
...
 
-For fetching and uploading data, our workflow requires S3 credentials and variables. These credentials will be provided as kubernetes secrets, and the variables will be passed into the workflow. Modify the below values to suit your environment.
+NOTE: replace `192.168.100.155` with your own Minio server address that setup in step 4.
 
 ```
-export S3_ENDPOINT=s3.us-west-2.amazonaws.com  #replace with your s3 endpoint in a host:port format, e.g. minio:9000
-export AWS_ENDPOINT_URL=https://${S3_ENDPOINT} #use http instead of https for default minio installs
-export AWS_ACCESS_KEY_ID=xxxxx
-export AWS_SECRET_ACCESS_KEY=xxxxx
+export S3_ENDPOINT=192.168.100.155:9000
```
elsonrodriguez: This makes a lot of assumptions about the network configuration.

mnist/README.md
elsonrodriguez: @jlewi The image was being built by the reader in the previous version, the instructions on how to build/push were deleted in this update.

mnist/model-train.yaml
```diff
@@ -36,10 +36,6 @@ spec:
       value: true
     - name: model-serving-servicetype
       value: ClusterIP
-    - name: model-serving-ks-url
```
elsonrodriguez: This pins the version so that the examples continue to work despite any changes in master. Otherwise the example can break depending on code changes.

elsonrodriguez:
@chenzhiwei It looks like a lot of testing/validation went into creating this PR, so I hate to say that it needs to be reworked.

The scope is way beyond intend purpose (updating the example so it works with newer versions of kubeflow), and some of the changes are breaking (hard-coded IPs, unpinning kubeflow versions, removal of azure instructions) or questionable (defaulting to minio, adding a web client)

I think a new PR should be opened with just the version fixes in place, and the other suggestions (like the web client, or flow changes) should be tackled separately.

@jlewi Regarding the Argo bits, the intention originally was to demonstrate what an automated ML pipeline might look like. However while I think there's value in that, it is also pretty steep for a newcomer to this project to have to grok Argo before doing some mnist. I'd endorse an attempt at taking the argo bits out and letting the user do everything manually, while also preserving the argo bits somewhere to give automation folks something to work with.

dsdinter:
> Thanks again so much for taking the time to cleanup and fix this example!
> 
> A high level question for you; what do you think we should highlight and try to show in this example?
> 
> Should the goal of this example be to have a simple example that shows how to train and deploy a model using Kubeflow?
> 
> Should we keep the Argo bits? The Argo workflow right now is creating a ksonnet app and adding TFServing to it. What do you think of this pattern? I think the more common case would be people would have a ksonnet application checked in somewhere and then just apply the components in that app; rather than creating an app at run time.
> 
> I think the pattern we'd probably recommend is using GitOps
> https://www.kubeflow.org/docs/guides/gitops-for-kubeflow/
> 
> If we remove the Argo bits; that might simplify the example and make it more suited as a getting started example for users.
> 
> @dsdinter WDYT? You are basically working on a similar example for PyTorch: #274 ?
> 
> /cc @texasmichelle

Thanks, I agree about keeping the examples without the Argo workflows to keep them simple (At least with less moving parts).
For my example I created a ksonnet application and the components, then removed my test environment to avoid keeping my own residual tests.
We can then have another folder that keeps all Argo workflows about the examples consolidated, which can then be used to run our E2E tests? Pretty sure we can reuse the workflows hence suggesting a common folder, that will enforce some consistency about workflows across all examples.

mnist/README.md
chenzhiwei: Hi, as @jlewi suggested, put things into right place. I'd like to put the azure instruction in kubeflow getting started document.

mnist/README.md
chenzhiwei: this numerical format is very helpful for beginners, they just need to follow the steps one by one and finally get a good result.

Any suggestions on this? @jlewi

mnist/README.md
chenzhiwei: Firstly, the previous example is using S3. Secondly, creating PV seems more complex than using Minio especially when running multiple mnist models.

mnist/README.md
chenzhiwei: Running in the cluster need PV to persist the data, usually in demo environment there is no out-of-box PV.

This guide is for newbies, I want to ensure they can follow the steps one by one to run this example. For skilled person, they already know how to tweak the steps to use their own solution, such as they may you Minio helm chart to deploy their own Minio.

mnist/README.md
chenzhiwei: There is already a dockerfile directory inside this example, and the steps are located there.

Later when I move the dockerfile to the location @jlewi suggested, I will update it to reference the link.

mnist/README.md
chenzhiwei: it is already in the example, I did not change it, so there is not diff in this PR. https://github.com/chenzhiwei/kubeflow-examples/blob/mnist-model/mnist/tf-user.yaml

mnist/README.md
chenzhiwei: in line 78, there is a description about this.

mnist/README.md
chenzhiwei: this example already has this feature and I think it is reasonable to separate the train and serving model. I did not change it so there is not diff.

mnist/README.md
chenzhiwei: the server is the node that deployed k8s cluster, they need download this example to server. I think all people can know this without tips.

mnist/dockerfile/Dockerfile.ksonnet
chenzhiwei: yes, the model-train.yaml is a workflow, the ks is run inside the workflow.

mnist/dockerfile/README.md
chenzhiwei: Created one: https://github.com/kubeflow/examples/issues/319 and will link too this README.

mnist/dockerfile/ksonnet-entrypoint.sh
chenzhiwei: actually we don't need this. the env var is set in pod spec.

and this line does not use `export`, the actually command will not recognize this. I leave this unchanged in this PR.

mnist/mnist-webapp/README.md
chenzhiwei: there is no docker container, but your suggestion is good. I can build a minist-webapp container and then remove this mnist-webapp code.

Thank you very much.

mnist/model-train.yaml
chenzhiwei: the model-train.yaml is argo workflow spec file.

mnist/model-train.yaml
chenzhiwei: I build the example code to docker image, so when serving the model, it will not call github api to fetch code, this can solve the github api limit issue.

Build-in the code(54KB) to docker image can also helps in offline deploy mode.

mnist/README.md
jlewi: I'm a bit lost. Isn't the server deployed inside Kubernetes?

mnist/README.md
chenzhiwei: The example images will be published in kubeflow which will be addressed in this separate issue: https://github.com/kubeflow/examples/issues/319

The instructions may not work well in future, the packages changes a lot, the only way to make it work all the time is pre-build it and publish to kubeflow.

Users just need to get it and use it, most users will not build the images by themself, if they want they can build them, later I will update the details in issue: https://github.com/kubeflow/examples/issues/319

This PR is too large, I want to reduce the changes and make it merged fast.

mnist/README.md
chenzhiwei: I will remove this webapp, and submit new PR later to include this webapp. Reduce the PR size.

jlewi:
Sorry for the slow reply I'm at Kubecon this week so I haven't had much time. Will try to review this as soon as i can.

@dsdinter Any interest in looking at this in the meantime?

mnist/README.md
jlewi: Ok thanks.

mnist/README.md
jlewi: I see thanks.

mnist/README.md
jlewi: How does Minio lead to better support for multiple mnist models?

Isn't the point of Minio to allow people to run an Object store on prem?

mnist/README.md
jlewi: I agree we should target newbies. So why introduce MinIO at all? If folks are running on a cloud they should be able to just use S3,GCS or HDFS directly (TF natively supports those filesystems).

If folks are running on prem and don't have an object store, why not just have them run non-distributed?

mnist/README.md
jlewi: Ok thanks.

mnist/mnist-webapp/README.md
jlewi: Thanks.

mnist/README.md
jlewi: Good to know thanks.

mnist/README.md
jlewi: Can we just make installing the preqruisites step #1? 

Also can you add a space between the "." and the text.

mnist/README.md
jlewi: +1

mnist/README.md
jlewi: +1 to @elsonrodriguez to not recommending minio for beginners. That seems like a distraction. If users don't have a Cloud account they are probably running on minikube/microk8s/docker. So I think the best option is to just use a PV even if that means running non-distributed.

chenzhiwei:
Actually I want to create a new mnist example at first, but I don't want make duplicated work, so I chose a harder way which is creating this PR.

Do you agree to create a new mnist example with the new versions and remove the argo(I agree that argo makes things complex)?

jlewi:
I'd like a single mnist example. Whether we get there by "refactoring" the existing sample; or deleting the current one and creating a new one; doesn't really matter to me. 

I support the idea of removing Argo.

texasmichelle:
There's a [codelab](https://codelabs.developers.google.com/codelabs/kubeflow-introduction/index.html?index=..%2F..index#0) that uses the existing example, but as long as that gets updated simultaneously I'm fine with replacing. I agree that a single mnist example makes the most sense.

@chenzhiwei could you either update or close this PR? Breaking it down into a number of smaller issues/PRs is always an option.

chenzhiwei:
sorry, I have no time to work on this now.

PR #320, opened 2018-11-09
Modify K8s models to export the models; tensorboard manifests

jlewi:
* Use a K8s job not a TFJob to export the model.
* Start an experiments.libsonnet file to define groups of parameters for
  different experiments that should be reused

* Need to install tensorflow_hub in the Docker image because it is
  required by t2t exporter.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/320)

### DIFF
--- a/code_search/kubeflow/components/experiments.libsonnet
+++ b/code_search/kubeflow/components/experiments.libsonnet
@@ -1,0 +1,11 @@
+// Data for various experiments.
+// Paths are deliberately hard coded so they get versioned and checked into source control.
+{
+  "demo-trainer-11-07-dist-sync-gpu": {
+    name: "demo-trainer-11-07-dist-sync-gpu",
+    outputDir: "gs://code-search-demo/models/20181107-dist-sync-gpu",
+    train_steps: 200000,
+    eval_steps: 100,
+    hparams_set: "transformer_base",
+  },
+}
--- a/code_search/kubeflow/components/export-model.libsonnet
+++ b/code_search/kubeflow/components/export-model.libsonnet
@@ -1,0 +1,68 @@
+{
+  parts(params, env):: {
+    job: {
+      apiVersion: "batch/v1",
+      kind: "Job",
+      metadata: {
+        name: params.name,
+        namespace: env.namespace,
+        labels: {
+          app: params.name,
+        },
+      },
+      spec: {
+        replicas: 1,
+        template: {
+          metadata: {
+            labels: {
+              app: params.name,
+            },
+          },
+          spec: {
+            // Don't restart because all the job should do is launch the Dataflow job.
+            restartPolicy: "Never",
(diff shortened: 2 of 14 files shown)


code_search/kubeflow/components/export-model.libsonnet
```diff
@@ -0,0 +1,68 @@
...
+            containers: [
+              {
+                name: "exporter",
+                image: params.image,
+                command: [
+                  "t2t-exporter",
+                  "--problem=" + params.problem,
+                  "--data_dir=" + params.dataDir,
+                  // TODO(jlewi): If we specify checkpoint path then we can specify a particular
+                  // checkpoint. Does this then allow us to specify a different directory for outputDir?
```
cwbeitel: I think with t2t-exporter you can specify the export dir via --export or --export_dir instead of it going to the default FLAGS.output_dir/export, if that's what you were getting at.

code_search/kubeflow/components/export-model.libsonnet
```diff
@@ -0,0 +1,68 @@
...
+      spec: {
+        replicas: 1,
+        template: {
+          metadata: {
+            labels: {
+              app: params.name,
+            },
+          },
+          spec: {
+            // Don't restart because all the job should do is launch the Dataflow job.
```
cwbeitel: rm?

code_search/kubeflow/components/experiments.libsonnet
```diff
@@ -0,0 +1,11 @@
+// Data for various experiments.
+// Paths are deliberately hard coded so they get versioned and checked into source control.
```
cwbeitel: I really like this idea in combination with being able to construct a configuration in python, from a notebook.

code_search/kubeflow/components/export-model.libsonnet
jlewi: Thanks @cwbeitel 
Filed /kubeflow/examples#331

I'll fix it in a follow on PR because I want to get this PR submitted to unblock @IronPan

jlewi:
@cwbeitel If you like the idea of Python you should check out kubeflow/pipelines. That uses a DSL in python to describe workflows.

Whether its better to use Python or ksonnet is a complicated question and probably comes down to what you prefer.

ksonnet/jsonnet are often preferred for describing infrastructure because its harder to do "bad things". Bad things would be things like having a side effect (e.g. modifying infrastructure or writing to a DB) just to generate the spec for the infrastructure.

jlewi:
@cwbeitel Also its possible to write tests for ksonnet. And we should do that.

jlewi:
@IronPan Can you please re-LGTM; I made some trivial changes.

IronPan:
Thanks Jeremy for the change.

PR #323, opened 2018-11-10
Add preprocess step container

IronPan:
Wraps the preprocess step into a container. 
Also rename flag wait_until_finish -> wait_until_finished since wait_until_finish seems collide with some flag with dataflow runner 

TODO: The embedding step might be able to reuse the image since the only difference is the entrypoint.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/323)


jlewi:
Can we use the existing Dataflow docker image?
https://github.com/kubeflow/examples/blob/master/code_search/docker/t2t/Dockerfile.dataflow

We already use this Docker image to submit the Dataflow jobs. So can we reuse this image as opposed to building a new image?

Does Dataflow's built-in flag "wait_until_finish" flag do what we want? Can we get rid of our custom flag?

How will pod failures be handled? A Dataflow job could be long running; if the pod gets restarted will we continue to wait for the already existing job to finish or will we submit a new job?

code_search/pipeline/components/preprocess/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN mkdir -p /usr/local/gcloud \
+  && tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
+  && /usr/local/gcloud/google-cloud-sdk/install.sh
+
+# Adding the package path to local
+ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
+
+WORKDIR examples/code_search/src
+
+ENTRYPOINT ["/bin/bash", "-c", "gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS} && python2 -m code_search.dataflow.cli.preprocess_github_dataset"]
```
jlewi: I don't think running gcloud auth activate-service-account;
We already have a K8s job running preprocess_github_dataset to submit the Workflow. And this works without running activate-service-account.

IronPan:
> Can we use the existing Dataflow docker image?
https://github.com/kubeflow/examples/blob/master/code_search/docker/t2t/Dockerfile.dataflow
Yes switch to use it in https://github.com/kubeflow/examples/pull/334. I'll close this one

> Does Dataflow's built-in flag "wait_until_finish" flag do what we want? Can we get rid of our custom flag?
I'll investigate into that a little bit. 

> How will pod failures be handled? A Dataflow job could be long running; if the pod gets restarted will we continue to wait for the already existing job to finish or will we submit a new job?
Argo won't restart the pod so the pipeline job will fail. If the user restart the pipeline a new job will be created. This is how current pipeline dataflow [components](https://github.com/kubeflow/pipelines/tree/master/components/dataflow) works. A proper error handling requires proper caching, which is the in the scope of pipeline's future plan.

code_search/pipeline/components/preprocess/Dockerfile
IronPan: ack. this is removed in #334

PR #324, opened 2018-11-10
Create a component to submit the Dataflow job to compute embeddings for code search

jlewi:
* Update Beam to 2.8.0
* Remove nmslib from Apache beam requirements.txt; its not needed and appears
  to have problems installing on the Dataflow workers.

* Spacy download was failing on Dataflow workers; reinstalling the spacy
  package as a pip package appears to fix this.

* Fix some bugs in the workflow for building the Docker images.

* Split requirements.txt into separate files for the UI and for Dataflow workers

  * On Dataflow workers nmslib install is failing. The Dataflow workers also don't need nmslib.
  * On Dataflow workers we want to avoid installing unnecessary dependencies because it slows down worker startup.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/324)


jlewi:
#320 should be submitted first.

code_search/src/requirements.txt
```diff
@@ -1,10 +1,13 @@
 astor~=0.7.0
-apache-beam[gcp]~=2.6.0
+apache-beam[gcp]~=2.8.0
 Flask~=1.0.0
 nltk~=3.3.0
-nmslib~=1.7.0
+# TODO(jlewi): nmslib builds are failing on Dataflow workers with Apache beam 2.8.0.
+# We shouldn't need nmslib in the Dataflow jobs.
```
cwbeitel: Right, we shouldn't need nmslib in computing function embeddings. Presumably you've confirmed it works without it (noting is trying to import nmslib). In which case this can be removed.

jlewi: Done.

jlewi:
This is now ready for review.

/assign @IronPan

code_search/kubeflow/components/params.libsonnet
```diff
@@ -20,7 +20,7 @@
       eval_steps: 10,
       image: "gcr.io/kubeflow-examples/code-search:" + imageTag,
       imageGpu: "gcr.io/kubeflow-examples/code-search-gpu:" + imageTag,
-      dataflowImage: "gcr.io/kubeflow-examples/code-search-dataflow:v20181106-v0.2-76-g611636c-dirty-860631",
+      dataflowImage: "gcr.io/kubeflow-examples/code-search-dataflow:v20181109-dc79384",
```
IronPan: Did you build the image and push manually? Does kubeflow has release process to automatically update image tag?

(For /example project this might not matters that much, but just curious because /pipeline project faces this issue)

jlewi: In this particular case I built the image manually. Its possible to build images automatically by taking advantage of our CI/CD system.

PR #330, opened 2018-11-11
Change dataflow default workdir to /src

IronPan:
Otherwise when I want to execute dataflow code 
```
python2 -m code_search.dataflow.cli.create_function_embeddings \
```
it complains no setup.py

I could workaround by using workingdir option to switch the dir when launching the container, but setting it to default would be more convenient.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/330)

### DIFF
--- a/code_search/docker/t2t/Dockerfile.dataflow
+++ b/code_search/docker/t2t/Dockerfile.dataflow
@@ -17,6 +17,6 @@ RUN python -m spacy download en
 ADD src/code_search /app/code_search
 ADD src             /src
 
-WORKDIR /app
+WORKDIR /src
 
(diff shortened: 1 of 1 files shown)


IronPan:
In addition, this would be very useful for resuing the image to build a kf pipeline. See limitation of current pipeline.
https://github.com/kubeflow/pipelines/issues/204

jlewi:
FYI; You can also set the workdir in the command.
https://github.com/kubeflow/examples/blob/2669a54cce5cadb463b920bdd3429d85ad12df75/code_search/kubeflow/components/export-model.libsonnet#L46

PR #333, opened 2018-11-12
Remove ksonnet registry from dockerignore file

IronPan:
In order to build a pipeline that can runs ksonnet command, the ksonnet registry need to be containerized.
Remove it from dockerignore to unblock the work.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/333)

### DIFF
--- a/code_search/.dockerignore
+++ b/code_search/.dockerignore
@@ -1,4 +1,3 @@
 **/.git
 **/.DS_Store
 **/node_modules
-**/kubeflow


jlewi:
So is the idea to bake the ksonnet app into the docker image and then apply it?

An alternative would be to have the pipeline step checkout the ksonnet app and then apply it?

Any thoughts about which of these approaches is better?

IronPan:
Having it baked in gives quicker iteration. People don't need to maintain a separate dockerfile to grab local change and test it out. Also it's lighter in terms of release process. One can just publish the image and have it tested and done.
Otherwise the release process would be 1. cut a release from github 2. build a image using that release number and test it out

Long term wise, I feel we should have dedicated DSL operator for tfjob/job/tfserving etc. and we don't have to use ks for deployment. So this won't be needed anymore.

IronPan:
Due to the limitation of the current DSL operator, I won't be able to use them. Here is an issue for tracking. 
https://github.com/kubeflow/pipelines/issues/218

PR #334, opened 2018-11-12
[WIP] Convert code search sample into a KF pipeline

IronPan:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/334)


code_search/src/code_search/dataflow/cli/arguments.py
```diff
@@ -29,7 +29,7 @@ def add_parser_arguments(parser):
                       help='BigQuery dataset for output results')
   additional_args_parser.add_argument('--pre_transformed', action='store_true',
                       help='Use a pre-transformed BigQuery dataset')
-  additional_args_parser.add_argument('--wait_until_finish', action='store_true',
+  additional_args_parser.add_argument('--wait_until_finished', action='store_true',
```
IronPan: Comment from #323
Does Dataflow's built-in flag "wait_until_finish" flag do what we want? Can we get rid of our custom flag?

jlewi: Do you know the answer?

jlewi:
Can you update the PR description to provide more detail about what this PR is doing?

It looks like this is adding a Docker container and script that uses ksonnet to deploy the search index creator?

jlewi:
What are the remaining steps to make this PR ready?

code_search/docker/ks/deploy.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# Default namespace to deploy the ksonnet component to
+NAMESPACE="kubeflow"
+# Providing negative value to kubeflow wait would wait for a week
+TIMEOUT="-1s"
+
+usage()
+{
+    echo "usage: deploy.sh
+    [--working_dir The base GCS bucket ]
```
jlewi: Why is working_dir just a GCS bucket and not a full path?
Should there be an argument specifying where the index is created so that we don't override existing indexes?

jlewi:
FYI; I filed an issue for this #349 with more detail.

PR #338, opened 2018-11-17
Fix the K8s job to create the nmslib index.

jlewi:
* Install nmslib in the Dataflow container so its suitable for running
  the index creation job.

* Use command not args in the job specs.

* Dockerfile.dataflow should install nmslib so that we can use that Docker
  image to create the index.

* build.jsonnet should tag images as latest. We will use this to use
  the latest images as a layer cache to speed up builds.

* Set logging level to info for start_search_server.py and
  create_search_index.py

* Create search index pod keeps was getting evicted because node runs out of
  memory

* Add a new node pool consisting of n1-standard-32 nodes to the demo cluster.
 These have 120 GB of RAM compared to 30GB in our default pool of n1-standard-8

* Set requests and limits on the creator search index pod.

* Move all the config for the search-index-creator job into the
  search-index-creator.jsonnet file. We need to customize the memory resources
  so there's not much value to try to sharing config with other components.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/338)


code_search/kubeflow/components/search-index-creator.jsonnet
```diff
@@ -3,5 +3,82 @@ local nms = import "nms.libsonnet";
...
+		          "--index_file=" + params.indexFile,
+		        ],
+            env: [
+              {
+                name: "GOOGLE_APPLICATION_CREDENTIALS",
+                value: "/secret/gcp-credentials/user-gcp-sa.json",
+              },
+            ],
+            // Creating the index requires a lot of memory.
+            resources: {
```
IronPan: not sure if i follow here. the actual workload is send to dataflow. why the job's pod itself need much resource?

PR #342, opened 2018-11-20
[mnist_pytorch] fix train image

dsdinter:
Checkout Pytorch 1.0rc1 release as latest master seems to have MPI backend detection broken

### DIFF
--- a/pytorch_mnist/serving/seldon-wrapper/mnistddpserving.py
+++ b/pytorch_mnist/serving/seldon-wrapper/mnistddpserving.py
@@ -48,8 +48,8 @@ class mnistddpserving():
   def __init__(self):
     self.class_names = ["class:{}".format(str(i)) for i in range(10)]
     self.model = Net()
-    # TODO parametrise path to load model
-    self.model.load_state_dict(torch.load("/mnt/kubeflow-gcfs/pytorch/model/model.dat",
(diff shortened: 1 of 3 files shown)


dsdinter:
@jlewi please review, I will start working on a new PR to enable tests in case anything breaks again.

.pylintrc
```diff
@@ -13,7 +13,7 @@ ignore=third_party
 
 # Add files or directories matching the regex patterns to the blacklist. The
 # regex matches against base names, not paths.
-ignore-patterns=prediction_pb2.py,prediction_pb2_grpc.py,mnist_DDP.py
+ignore-patterns=prediction_pb2.py,prediction_pb2_grpc.py,mnist_DDP.py,mnistddpserving.py
```
jlewi: Why ignore mnistddpserving.py? Is this code copied from somewhere else?

jlewi:
Thanks
/lgtm
/approve

.pylintrc
dsdinter: Thanks - Not really, I was getting errors with the method forward() as pylint expected an arbitrary argument list in the method signature def forward(self, *x), I will try to fix the method signature to make it work in both the trainers and serving modules in the next release (Original examples from where I took code snippets have same problem).
Pylint error "Parameters differ from overridden 'forward' method (37:2) [arguments-differ]"

Issue #346, opened 2018-11-22
[object_detection] lack ks environment directory


hougangliu:
In object_detection ks_app, there is no environments directory, which causes `ks apply ${ENV}` fail.
Also, we need guide user to configure env correctly.

jlewi:
Don't users need to create their own environment directory corresponding to their specific  cluster?

PR #347, opened 2018-11-22
Add ks env configuration(#346)

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/347)

### DIFF
--- a/object_detection/ks-app/environments/base.libsonnet
+++ b/object_detection/ks-app/environments/base.libsonnet
@@ -1,0 +1,4 @@
+local components = std.extVar("__ksonnet/components");
+components + {
+  // Insert user-specified overrides here.
+}
--- a/object_detection/setup.md
+++ b/object_detection/setup.md
@@ -42,6 +42,9 @@ Let's make use of the app to continue with the tutorial.
 
 ```
 cd ks-app
+ENV=default
+ks env add ${ENV} --context=`kubectl config current-context`
+ks env set ${ENV} --namespace kubeflow
 ```
 
 ## Preparing the training data
--- a/object_detection/ks-app/app.yaml
+++ b/object_detection/ks-app/app.yaml
@@ -1,17 +1,4 @@
 apiVersion: 0.1.0
-environments:
-  default:
-    destination:
-      namespace: kubeflow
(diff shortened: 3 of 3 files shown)


object_detection/setup.md
```diff
@@ -42,6 +42,10 @@ Let's make use of the app to continue with the tutorial.
 
 ```
 cd ks-app
+ks env rm default
+ENV=default
+ks env add ${ENV} --context=`kubectl config current-context`
```
jlewi: Why are you explicitly setting context to the current-context? Won't ks do this automatically?

jlewi:
Thanks for the fix!

Should we avoid checking in the default environment?

So was the problem that we were missing the file base.libsonnet so as a result `ks env add`` wasn't working?

object_detection/setup.md
hougangliu: ks do this only when run "ks init" when create ks_app.

hougangliu:
in a ks app, environments directory is in need. when we run "ks env add", a new sub-directory with env name will be created, and main.jsonnet import base.libsonnet (base.libsonnet is created when run "ks init") by default.

this patch aims to guide newcomer of ks or kubeflow can walk through this example with little effort (he only need copy &paste)

object_detection/setup.md
hougangliu: yes, agree with you. We don't need create an existing env. Updated.

hougangliu:
Deleted default env, user can add his new env.

object_detection/ks-app/environments/base.libsonnet
```diff
@@ -0,0 +1,4 @@
+local components = std.extVar("__ksonnet/components");
+components + {
+  // Insert user-specified overrides here.
```
gyliu513: Instead of only adding comments here, how about adding an example here?

like 
```
// Insert user-specified overrides here.
// user override data???
```

hougangliu: actually the comment is automatically generated by "ks init"

gyliu513: Ah, I see, thanks.

Issue #349, opened 2018-11-22
[code_search] Create a pipeline to generate and update the index of embeddings


jlewi:
We'd like to create a pipeline to update the code embeddings with the latest data in GitHub.

The pipeline should consist of three steps

1. Run the Dataflow job to compute embeddings
  
   * ksonnet componet is 
https://github.com/kubeflow/examples/blob/master/code_search/kubeflow/components/submit-code-embeddings-job.jsonnet

   * We should give the output a unique directory

1. Run the K8s job to convert csv files to nmslib index

    * https://github.com/kubeflow/examples/blob/master/code_search/kubeflow/components/search-index-creator.jsonnet

1. Push a commit/or create a PR to update the nmslib index.

    * #348 is updating the ksonnet component to deploy the UI
    * What we want is to run a script that will change the value of the arguments lookupFile and indexFile
  *https://github.com/kubeflow/examples/blob/ab35c85fd91e52b87cc4068fa4bd389a783665a1/code_search/kubeflow/components/search-index-server.jsonnet#L48

texasmichelle:
Has this issue been resolved?

PR #352, opened 2018-11-23
Create a script to update the index and lookup file used to serve predictions

jlewi:
* This script will be the last step in a pipeline to continuously update
  the index for serving.

* The script updates the parameters of the search index server to point
  to the supplied index files. It then commits them and creates a PR
  to push those commits.

* Restructure the parameters for the search index server so that we can use
  ks param set to override the indexFile and lookupFile.

* We do this because we want to be able to push a new index by doing
  ks param set in a continuously running pipeline
* Remove default parameters from search-index-server

* Create a dockerfile suitable for running this script.

Related to #349

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/352)


jlewi:
#348 needs to be submitted first

/assign @IronPan

jlewi:
#348 is submitted so this is ready for review.

code_search/docker/index_updater/update_index.sh
```diff
@@ -0,0 +1,95 @@
...
+
+This PR is automatically generated by update_index.sh.
+
+This PR updates the index and lookup file used to serve
+predictions.
+EOF
+
+# Create a pull request
+if (! ${dryrun}); then
+hub pull-request --base=${base} -F ${FILE}
```
IronPan: Maybe adding a mode to merge the change without sending a PR? It would be useful for automated pipeline

PR #359, opened 2018-11-25
Setup ArgoCD to synchornize the code search web app with the demo cluster

jlewi:
* Follow argocd instructions
  https://github.com/argoproj/argo-cd/blob/master/docs/getting_started.md
  to install ArgoCD on the cluster

  * Down the argocd manifest and update the namespace to argocd.
  * Check it in so ArgoCD can be deployed declaratively.

* Update README.md with the instructions for deploying ArgoCD.

Move the web app components into their own ksonnet app.

* We do this because we want to be able to sync the web app components using
  Argo CD

* ArgoCD doesn't allow us to apply autosync with granularity less than the
  app. We don't want to sync any of the components except the servers.

* Rename the t2t-code-search-serving component to query-embed-server because
  this is more descriptive.

* Check in a YAML spec defining the ksonnet application for the web UI.

Update the instructions in nodebook code-search.ipynb

  * Provided updated instructions for deploying the web app due the
  fact that the web app is now a separate component.

  * Improve code-search.ipynb
    * Use gcloud to get sensible defaults for parameters like the project.
    * Provide more information about what the variables mean.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/359)


code_search/code-search.ipynb
```diff
@@ -273,13 +279,78 @@
...
+    "    * **lookupFile**: set this to the GCS location of the CSV produced by the job to create the nmslib index of the embeddings for all GitHub data\n",
+    "    * **indexFile*:* set this to the GCS location of the nmslib index for all the data in GitHub\n",
+    "\n",
+    "Configure your ksonnet environment to use your experiment\n",
+    "\n",
+    "* Open **kubeflow/environments/${ENVIRONMENT}/globals.libsonnet**\n",
+    "* Define the following global parameters\n",
+    "\n",
+    "  * **experiment**: Name of your experiment; should correspond to a key defined in **experiments.libsonnet**\n",
+    "  * **project**: Set this to the GCP project you can use\n",
```
IronPan: I am wondering why not move all param to experiments.libsonnet, except experiment name ?
the project argument for example is duplicated in two places.

code_search/code-search.ipynb
```diff
@@ -196,7 +202,7 @@
     "cd kubeflow\n",
     "\n",
     "# Update Ksonnet to point to the Kubernetes Cluster\n",
-    "ks env add code-search --context $(kubectl config current-context)\n",
+    "ks env add code-search\n",
     "\n",
     "# Update Ksonnet to use the namespace where kubeflow is deployed. By default it's 'kubeflow'\n",
     "ks env set code-search --namespace=kubeflow\n",
```
IronPan: maybe define a env variable for the environment and used for all places inside this guideline?

code_search/ks-web-app/app.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: 0.2.0
+environments:
+  cs_demo:
+    destination:
+      namespace: cs-web-app
+      server: https://35.185.115.154
```
IronPan: i see similar IP in a few places? was this intended? to me https://kubernetes.default always works.

code_search/code-search.ipynb
jlewi: I think that's where we're headed: kubeflow/examples#308
This PR was already quite big so I didn't want to try to clean everything up in this PR.

I'm not sure whether it makes sense to move everything into experiments or define some other things as globals.

code_search/code-search.ipynb
jlewi: I think that makes sense but I figured I'd leave that for a follow on PR; since the notebook wasn't using an env variable.

code_search/ks-web-app/app.yaml
jlewi: kubernetes.default should only work if you are running in cluster.

But we want our ksonnet to be able to work off cluster. In which case we need to specify the actual IP of the master.

PR #361, opened 2018-11-26
Add update search index pipeline

IronPan:
Add a kubeflow pipeline to update search index. 

Currently the pipeline does two things
- Create a dataflow job to compute function embedding
- Create a K8s job to create search index

The compute function embedding op is a container that wraps the direct python call to dispatch dataflow job. (TBD switch to use ksonnet component)
The create search index op is a container that wraps the existing create search index ksonnet component.

See README in pipeline/ dir for more details how to run the pipeline

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/361)


code_search/Makefile
```diff
@@ -46,7 +46,14 @@ build-dataflow:
 			 .
 	@echo Built $(IMG)-dataflow:$(TAG)
 
-build: build-cpu build-gpu build-dataflow
+build-search-index-creator:
```
jlewi: Why do we need a new container? Why can't we use the existing Dataflow container?

code_search/docker/search-index-creator/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM debian:7
```
jlewi: Can't we just use: https://github.com/kubeflow/examples/blob/master/code_search/docker/t2t/Dockerfile.dataflow?
And add to it if needed?

code_search/docker/search-index-creator/deploy.sh
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
```
jlewi: Add a comment explaining what this script is for.

code_search/docker/search-index-creator/deploy.sh
```diff
@@ -0,0 +1,90 @@
...
+kubectl config use-context "${KUBECTL_CONTEXT_NAME}"
+ks env set "${KS_ENV_NAME}" --namespace="${NAMESPACE}"
+
+# Apply parameters
+ks param set ${COMPONENT} dataDir ${DATA_DIR} --env ${KS_ENV_NAME}
+ks param set ${COMPONENT} jobNameSuffix ${WORKFLOW_ID} --env ${KS_ENV_NAME}
+ks param set ${COMPONENT} lookupFile ${WORKING_DIR}/code-embeddings-index/${WORKFLOW_ID}/embedding-to-info.csv --env ${KS_ENV_NAME}
+ks param set ${COMPONENT} indexFile ${WORKING_DIR}/code-embeddings-index/${WORKFLOW_ID}/embeddings.index --env ${KS_ENV_NAME}
+
+echo "running ks apply pipeline -c ${COMPONENT}"
```
jlewi: You can probably delete this since "set -e" will echo every command run.

code_search/pipeline/index_update_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+      )
+  )
+
+def dataflow_function_embedding_op(project: 'GcpProject', runner: str, target_dataset: str, problem: str,
+                           data_dir: 'GcsUri',saved_model_dir: 'GcsUri',temp_location: 'GcsUri', staging_location: 'GcsUri',
+                           job_name: str, worker_machine_type: str,
+                           num_workers: int, step_name='dataflow_function_embedding'):
+    return default_gcp_op(
+        name = step_name,
+        image = 'gcr.io/yang-codesearch/code-search-dataflow:v20181124-23deb39',
```
jlewi: Can we use the same Dataflow container as used by the ksonnet component to submit the Dataflow job?

code_search/pipeline/index_update_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+        command = [
+            'python2',
+            '-m',
+            'code_search.dataflow.cli.create_function_embeddings',
+        ],
+        arguments = [
```
jlewi: Should we open an issue (per comment in PR description) to eventually switch to launching the ksonnet job to submit the Dataflow job?

I think the main advantage of using the ksonnet components is avoiding code duplication. If we reuse the ksonnet components then we know that the pipeline is launching the same jobs that are launched individually.

I think that would be useful for debugging/development because you could just do a one off run of a step by just running the component.

code_search/pipeline/index_update_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+            '--requirements_file', 'requirements.dataflow.txt',
+            '--wait_until_finished',
+        ]
+    )
+
+def search_index_creator_op(working_dir: str, data_dir: str, component: str, cluster_name: str, workflow_id: str):
+    return dsl.ContainerOp(
+        # use component name as step name
+        name = component,
+        image = 'gcr.io/yang-codesearch/code-search-search-index-creator:v20181126-6a63824',
```
jlewi: Same comment as above; can we reuse the Docker images in kubeflow-examples that the individual steps are already using?

code_search/pipeline/index_update_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+    temp_location = '%s/dataflow/%s/temp' % (working_dir, workflow_name)
+    staging_location = '%s/dataflow/%s/staging' % (working_dir, workflow_name)
+    function_embedding = dataflow_function_embedding_op(project, runner, target_dataset,problem,data_dir,saved_model_dir,
+                                        temp_location,staging_location,workflow_name,worker_machine_type,num_workers)
+    index_creator = search_index_creator_op(working_dir, data_dir, 'search-index-creator', cluster_name,
+                               workflow_name).after(function_embedding)
+
+
+if __name__ == '__main__':
+  import kfp.compiler as compiler
```
jlewi: Why is this imported here and not at the top?

Can you set the logging info like we do in other main functions
https://github.com/kubeflow/examples/blob/master/code_search/src/code_search/dataflow/cli/create_function_embeddings.py#L57

code_search/pipeline/index_update_pipeline.py
IronPan: ditto. I plan to start a new image that launch the ks component, and deprecate the old one if necessary

code_search/docker/search-index-creator/deploy.sh
```diff
@@ -0,0 +1,90 @@
...
+{
+    echo "usage: deploy.sh
+    [--working_dir The base GCS bucket ]
+    [--component What ksonnet component to run. Available components: search-index-creator ]
+    [--timeout How long to wait for the job to finish. e.g. 30m, 1h ]
+    [--namespace Namespace to deploy the ksonnet component to ]
+    [-h help]"
+}
+
+while [ "$1" != "" ]; do
```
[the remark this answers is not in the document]
IronPan: done

code_search/docker/search-index-creator/deploy.sh
IronPan: done

code_search/pipeline/index_update_pipeline.py
IronPan: done. https://github.com/kubeflow/examples/issues/363

code_search/pipeline/index_update_pipeline.py
IronPan: i've updated the dataflow image to reuse the existing one.

code_search/docker/parse_arguments.sh
```diff
@@ -0,0 +1,39 @@
...
+		value=true
+		eval ${name}="${value}"
+    else
+    	echo "Argument $1 did not match the pattern --{name}={value} or --{name}"
+    fi
+    shift
+  done
+}
+
+parseArgs $*
```
jlewi: Did you mean to copy code below line 27 here? Or does this code belong in the main file?

code_search/docker/index_updater/update_index.sh
```diff
@@ -14,33 +14,14 @@ set -ex
...
-
 usage() {
 	echo "Usage: update_index.sh --base=OWNER:branch --appDir=<ksonnet app dir> --env=<ksonnet environment> --indexFile=<index file> --lookupFile=<lookup file>"
 }
 
-parseArgs $*
+# List of required parameters
+names=(appDir env lookupFile indexFile base)
+
+source "../parse_arguments.sh"
```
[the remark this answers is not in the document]
IronPan: done

code_search/docker/ks/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM debian:7
+
+RUN apt-get update -q && apt-get upgrade -y && \
```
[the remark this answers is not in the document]
IronPan: done

code_search/docker/ks/search-index-creator.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
IronPan: done

code_search/docker/parse_arguments.sh
IronPan: yeah that was intentional. It's a common logic so i did as much as possible to avoid duplication. let me know if you want to move them to parent script.

code_search/pipeline/index_update_pipeline.py
jlewi: In general we should configure onfigure the logging module in main so that code can see logging to output info mesages.

PR #372, opened 2018-11-28
add pipeline environment to code search web app

IronPan:
This ensures when people fork the code, they can run pipeline directly without manually creating the environment first and check in.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/372)


jlewi:
How and where does this environment get used? This environment is using https://kubernetes.default so I think its expecting to be called from a pod running in the cluster.

Would it be better to call it "in-cluster" to capture the fact that must be executed from within cluster.

IronPan:
Yes this environment is used by the pipeline's update github step. https://github.com/kubeflow/examples/pull/373

The name is to align with here
https://github.com/kubeflow/examples/tree/master/code_search/kubeflow/environments/pipeline

But I'm have no strong preference on the naming. Please let me know if you think in-cluster is clearer and i can change both places

Issue #377, opened 2018-11-30
[object_detection] user has no idea about how to run inference


hougangliu:
In object_detection example, user can serve the model by doc step. But the doc doesn't mention how to run inference using the model. And it need take some time for the user to write a gRpc client.

PR #378, opened 2018-11-30
add object detection grpc client

hougangliu:
Fixes: #377

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/378)

### DIFF
--- a/object_detection/tf_serving_cpu.md
+++ b/object_detection/tf_serving_cpu.md
@@ -43,7 +43,46 @@ And you should see:
 E0621 19:20:34.134165172       7 ev_epoll1_linux.c:1051]     grpc epoll fd: 3
 2018-06-21 19:20:34.135354: I tensorflow_serving/model_servers/main.cc:288] Running ModelServer at 0.0.0.0:9000 ...
 ```
-Now you can use a gRPC client to run inference using your trained model!
+## Running inference using your model
+
+Now you can use a gRPC client to run inference using your trained model as below!
+
+First we need to install the dependencies (Ubuntu* 16.04 )
+```
+sudo apt-get install protobuf-compiler python-pil python-lxml python-tk
+pip install tensorflow
+pip install matplotlib
+pip install tensorflow-serving-api
+pip install numpy
+pip install grpcio
+```
+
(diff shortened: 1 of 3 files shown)


object_detection/ks-app/vendor/kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -25,7 +25,7 @@
     else
       $.params.httpProxyImage,
 
-    serviceType: "ClusterIP",
+    serviceType: "NodePort",
```
jlewi: Is there a reason you are making this override in the the vendor directory rather than modifying the generated component (in components) to override it.

object_detection/tf_serving_cpu.md
```diff
@@ -43,7 +43,43 @@ And you should see:
...
+protoc object_detection/protos/*.proto --python_out=.
+PYTHONPATH=:${TF_MODELS}/models/research:${TF_MODELS}/models/research/slim:${PYTHONPATH}
+```
+
+Now you can run the object detection client, and after that you should be seeing an image file in $OUT_DIR directory with the bounding boxes for detected objects.
+```
+#From examples/object_detection/serving_script directory
+OUT_DIR=`pwd`
+INPUT_IMG="image1.jpg"
+GRPC_PORT=`kubectl get svc pets-model -n kubeflow -o jsonpath={.spec.ports[0].nodePort}`
```
jlewi: Why not just use kubectl port-forward to open up a local port to the GRPC server?

Using a node port is usually insuficient because most people's clusters are behind a firewall so they can't directly connect to nodes. That's a good thing since we don't want folks opening ports to the internet by default.

object_detection/ks-app/vendor/kubeflow/tf-serving/tf-serving.libsonnet
hougangliu: Thanks. Remove this change.

object_detection/tf_serving_cpu.md
hougangliu: It make sense. Updated.

PR #380, opened 2018-11-30
Save batch-predict package in the docker images so that it can be used when launching Dataflow job on GCP

yixinshi:
Also create a separate Dockerfile for GPU only.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/380)


yixinshi:
Seems my git client is messed up. I will cancel this PR.

PR #383, opened 2018-11-30
Save the batch-predict package in the image; Create a separate Dockfile for GPU

yixinshi:
The package will be used when launching a Dataflow job on GCP


jlewi:
/assign @lluunn 
@lluunn Can you add yourself as a reviewer to the batch predict code?

Issue #381, opened 2018-11-30
[code_search] cleanup refactor bq_transform.BigQueryWrite


jlewi:
We should refactor BigQueryWrite
https://github.com/kubeflow/examples/blob/master/code_search/src/code_search/dataflow/transforms/bigquery.py#L42

Instead of creating a subclass of PTransform that wraps bea.io.WriteToBigQuery we should just create a wrapper class to construct a BigQuerySchema from a list of columns. then we can just define constants for the schemas and pass them to WriteToBigQuery

PR #388, opened 2018-12-01
Dataflow job should support writing embeddings to a different location (Fix #366).

### LINKED ISSUE #366: [code_search] Create code embeddings needs a way to write output to a different location
We need to be able to write the code embeddings to a different location.

We want to be able to run the Dataflow job multiple times each with a different output location; as opposed to overwriting the current outputs.

I think the current command line arguments make this difficult because the arguments don't distinguish between inputs & outputs so we can't change just the output location.

I tihnk the relevant parameters are

--target_dataset
--dataDir

I think target_dataset might be just for the output but we should double check

I think dataDir is used as an input (to locat the vocab) as well as for the output. We should split this into two parameters so we can put the output someplace lese.

jlewi:
* Dataflow job to compute code embeddings needs to have parameters controlling
  the location of the outputs independent of the inputs. Prior to this fix the
  same table in the dataset was always written and the files were always created
  in the data dir.

* This made it very difficult to rerun the embeddings job on the latest GitHub
  data (e.g to regularly update the code embeddings) without overwritting
  the current embeddings.

* Refactor the relevate  BQWriteTransforms (see #381)

  * Rather than create a wrapper class that bundles together a sink and schema
    we should have a separate helper class for creating BQ schemas and then
    use WriteToBigQuery directly.

  * Similarly for ReadTransforms we don't need a wrapper class that bundles
    a query and source. We can just create a class/constant to represent
    queries and pass them directly to the appropriate source.

* Change BQ write disposition to if empty so we don't overwrite existing data.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/388)

### DIFF
--- a/code_search/docker/t2t/Dockerfile.dataflow
+++ b/code_search/docker/t2t/Dockerfile.dataflow
@@ -22,6 +22,11 @@ RUN python -m spacy download en
 ADD src/code_search /app/code_search
 ADD src             /src
 
+# See: https://github.com/kubeflow/examples/issues/390
+# Dataflow will try to build a source package locally and we need
+# the path to match what we have in setup.py.
+RUN ln -sf /src/requirements.dataflow.txt /src/requirements.txt
+
 WORKDIR /src
 
 ENV PYTHONIOENCODING=utf-8 T2T_USR_DIR=/app/code_search/t2t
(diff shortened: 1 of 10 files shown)


jlewi:
/assign @IronPan 

@IronPan this is ready for review.

PR #394, opened 2018-12-03
Update the ks parameter 

IronPan:
- rename datadir to functionEmbeddingsDir for searchIndexCreator
- move tokenPairsBQTable functionEmbeddingsBQTable to common location. (The parameter in experiments.libsonnet can't be overriden by `ks param set`)

Also some other minor fixes

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/394)


code_search/kubeflow/components/submit-code-embeddings-job.libsonnet
```diff
@@ -30,10 +30,10 @@
 	              "python2",
 	              "-m",
 	              "code_search.dataflow.cli.create_function_embeddings",
-	              "--runner=DataflowRunner",	              
+	              "--runner=DataflowRunner",
 	              "--project=" + params.project,
-	              "--token_pairs_table=" + params.tokenPairsBQTable,
-	              "--function_embeddings_table=" + params.functionEmbeddingsBQTable,
+	              "--token_pairs_table=" + params.project + ":" + params.bqDataset + ".token_pairs",
+	              "--function_embeddings_table=" + params.project + ":" + params.bqDataset + ".code_embeddings_" + params.bqSuffix,
```
jlewi: Why did you make this change?  Why hardcode the fact that the token_pairs and function_embeddings_table always have these patterns?

In general when choosing a pattern like this e.g. the names of the tables we should do this at the highest level; i.e. have parameters that compeletly specify the value e.g. tokenPairsBQTable or functionEmbeddingsBQTable

This way its very easy to override or change the location of the table; the user doesn't need to reverse engineer the code to figure out which parameters to set to affect things.

code_search/kubeflow/components/submit-code-embeddings-job.jsonnet
```diff
@@ -8,6 +8,7 @@ local baseParams = std.extVar("__ksonnet/params").components["submit-code-embedd
 local experimentName = baseParams.experiment;
 local params = baseParams + experiments[experimentName] + {
   name: experimentName + "-embed-code",
+  bqSuffix: std.strReplace(self.jobNameSuffix, "-", "_"),
```
jlewi: See comment below. I think this should probably be done in experiments.libsonnet. For each experiment we should define the complete path/name of the things we care about (e.g. the individual BQTables). This makes the code more readable; i.e. they can look at experiments.libsonnet and see the full path of the BQ table for the function embeddings table.

code_search/kubeflow/components/submit-code-embeddings-job.libsonnet
IronPan: make sense. done.

code_search/kubeflow/components/submit-code-embeddings-job.jsonnet
IronPan: I didn't put the parameter for 'pipeline' env in experiments because it's not able to be overriden. I move them in param.libsonnet instead.

PR #396, opened 2018-12-03
Add preprocess github data step to code search pipeline

IronPan:
- Extend the function embedding step to read data from github public dataset if readGithubDatasetForFunctionEmbedding flag is true. 
- Refactor the preprocess code to pass table name with format PROJECT:DATASET.table instead of table, project and dataset name separately for conciseness. This aligns with rest of the code. 
- Expose failed_tokenize_table in preprocess as a parameter. 
- copy vocabularyFile to the data/ dir which is needed by function embedding step
- set backoffLimit for k8s jobs to 0 to stop rerun the job if failed. 
- Reformat the code to sort parameter alphabetically for easy maintenance.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/396)


jlewi:
nit: I'm a stickler for very descriptive PR descriptions. A good PR description would explain why we are making this change and provide an overview of the changes needed to accomplish that.

code_search/docker/ks/submit_code_embeddings_job.sh
```diff
@@ -21,37 +21,38 @@ workerMachineType=n1-highcpu-32
...
 	--dataDir=<data dir>
-	--functionEmbeddingsDir=<output function embedding dir>
-	--tokenPairsBQTable=<input token pairs BQ table>
 	--functionEmbeddingsBQTable=<output function embedding BQ table>
+	--functionEmbeddingsDir=<output function embedding dir>
+	--modelDir=<directory contains the model>
+	--namespace=<kubernetes namespace>
 	--numWorkers=<num of workers>
 	--project=<project>
+	--timeout=<timeout>
```
jlewi: Why do we need a timeout?

jlewi:
I thought the plan was going to be to extend the Dataflow job? I looked back at our conversation and maybe there was some confusion.

Here is where we define the Dataflow pipeline to compute the function embeddings
https://github.com/kubeflow/examples/blob/cea0ffde0d5c6d7275c93f1bf12eef038a8b9328/code_search/src/code_search/dataflow/cli/create_function_embeddings.py#L38

It starts with the token pairs.

So we can just change that code so that based on command line arguments we instead read the raw data directly from GitHub.

We basically need to add these lines at the Start of the Dataflow job

https://github.com/kubeflow/examples/blob/cea0ffde0d5c6d7275c93f1bf12eef038a8b9328/code_search/src/code_search/dataflow/cli/preprocess_github_dataset.py#L66

I think this will be better than adding a separate pipeline step.

1. Its much less code/complexity
1. It should be more efficient

    * By running the preprocessing and embedding steps in the same Dataflow job, we allow Dataflow to optimize the full pipeline; e.g. fuze preprocessing steps with batch inference steps.

code_search/docker/ks/submit_code_embeddings_job.sh
IronPan: this controls how long the argo step waits on the dataflow job.

code_search/docker/ks/Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM debian:7
+FROM google/cloud-sdk:225.0.0-slim
```
jlewi: Why do we need gcloud?

code_search/docker/ks/submit_code_embeddings_job.sh
jlewi: Why isn't the timeout in the Argo Step?

code_search/docker/ks/submit_code_embeddings_job.sh
```diff
@@ -21,37 +21,45 @@ workerMachineType=n1-highcpu-32
...
 }
 
 # List of required parameters
-names=(dataDir modelDir functionEmbeddingsDir tokenPairsBQTable functionEmbeddingsBQTable workingDir workflowId cluster namespace)
+names=(cluster dataDir failedTokenizeBQTable functionEmbeddingsBQTable functionEmbeddingsDir modelDir namespace project tokenPairsBQTable workflowId workingDir)
 
 source "${DIR}/parse_arguments.sh"
 source "${DIR}/initialize_kubectl.sh"
 
+gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS}
```
jlewi: Why do we need to copy the vocab file? Could we instead modify the Dataflow pipeline so that change isn't necessary?

code_search/kubeflow/components/submit-preprocess-job.jsonnet
```diff
@@ -1,75 +1,14 @@
-// Submit a Dataflow job to preprocess the input.
+// Submit a Dataflow job to preprocess the github public dataset
 local k = import "k.libsonnet";
 
+local experiments = import "experiments.libsonnet";
+local lib = import "submit-preprocess-job.libsonnet";
```
jlewi: What is the purpose of creating a submit-preprocess-job.libsonnet file?

code_search/kubeflow/environments/pipeline/globals.libsonnet
```diff
@@ -4,4 +4,5 @@
   // e.g. "dataflowImage", "trainerCpuImage", "trainerGpuImage",
   experiment: "pipeline",
   waitUntilFinish: "true",
+  readGithubDatasetForFunctionEmbedding: "true",
```
jlewi: Should this be a global? It only applies to a single component so shouldn't it be a component parameter?

code_search/src/code_search/dataflow/cli/arguments.py
```diff
@@ -42,6 +42,14 @@ def add_parser_arguments(parser):
                       help=('If specified read the entire GitHub dataset '
                             'specified as PROJECT:DATASET.TABLE. If not '
                             'specified we run a query to filter the data.'))
+  additional_args_parser.add_argument(
+                      '--read_github_dataset_for_function_embedding', action='store_true',
```
jlewi: Do we need the argument --read_github_dataset_for_function_embedding?

Could we infer based on other arguments? e.g. if we read the input from the token_pairs table we could have an argument for the token_pairs table. If that isn't set then we could read from the BigQuery GitHub dataset.

code_search/src/code_search/dataflow/cli/preprocess_github_dataset.py
```diff
@@ -59,8 +58,8 @@ def preprocess_github_dataset(argv=None):
       input_records = (pipeline
         | "Read Github Dataset" >> gh_bq.ReadGithubDataset(args.project))
     token_pairs = (input_records
-      | "Transform Github Dataset" >> github_dataset.TransformGithubDataset(args.project,
-                                                                            args.target_dataset)
+      | "Transform Github Dataset" >> github_dataset.TransformGithubDataset(
```
jlewi: We don't need to fix it in this PR since its a legacy of the existing code; but in general its better not to bundle sinks and sources in a transform. This makes the transform less reusable; e.g. you can't easily change the pipeline to write to a different type of sink or to do downstream processing on the results.

So a better pattern would be for TransformGitHubDataset to return token_pairs and failed_tokens and then here we would decide where to write them.

Its possible for a Transform to have multiple outputs. However, another way do this is to replace TransformGitHubDataset with a normal python function that would take input_records as an argument and return token_pairs, failed_tokenize_table.

jlewi:
Regarding data_dir I believe the only place it gets used as here

https://github.com/kubeflow/examples/blob/cea0ffde0d5c6d7275c93f1bf12eef038a8b9328/code_search/src/code_search/dataflow/transforms/function_embeddings.py#L30

Specifically the T2T library is using data_dir to fetch the vocab which is needed by the T2T library to vectorize the input before sending it to inference.

Is there an reason we we can't point data_dir at the directory where the vocab resides? i.e. why can't we use the same data_dir on each run of the Dataflow Job?

The Dataflow job shouldn't be writing anything to data_dir; (data_dir should only be used as an input not an input)

So can we avoid having to manually gsutil cp the vocab to a different location on each run?

code_search/docker/ks/submit_code_embeddings_job.sh
IronPan: It's not yet supported by DSL. Add issue to track the item
https://github.com/kubeflow/pipelines/issues/462

code_search/docker/ks/Dockerfile
IronPan: revert. it was used for copying vocabulary file

code_search/docker/ks/submit_code_embeddings_job.sh
IronPan: done. i instead pass an existing directory with vocabulary file as dataDir

code_search/kubeflow/components/submit-preprocess-job.jsonnet
IronPan: it's not necessary anymore. revert.

code_search/src/code_search/dataflow/cli/arguments.py
IronPan: done

code_search/kubeflow/environments/pipeline/globals.libsonnet
IronPan: revert this flag. thx

code_search/src/code_search/dataflow/cli/preprocess_github_dataset.py
IronPan: sounds good. just realized the two table parameters are optional. it will skip storing them if they are not specified. in this case, i'll skip specifying them.

code_search/src/code_search/dataflow/cli/arguments.py
jlewi: It looks like this parameter is still here?

code_search/src/code_search/dataflow/transforms/bigquery.py
```diff
@@ -77,12 +77,10 @@ class BigQueryWrite(beam.PTransform):
     ]
   """
 
-  def __init__(self, project, dataset, table, batch_size=500,
+  def __init__(self, table, batch_size=500,
```
jlewi: Per: https://github.com/kubeflow/examples/issues/381

Rather than make changes to this class; can we just update the calling locations to not use BigQueryWrite?
I will try to identify other change in this PR that motivated this change and provide a helpful comment there.

code_search/src/code_search/dataflow/transforms/github_dataset.py
```diff
@@ -44,10 +40,9 @@ def expand(self, input_or_inputs):
 
     pairs, tokenize_errors = tokenize_result.rows, tokenize_result.err
 
-    if self.target_dataset:
+    if self.failed_tokenize_table:
       (tokenize_errors  # pylint: disable=expression-not-assigned
-       | "Failed Tokenization" >> gh_bq.WriteFailedTokenizedData(self.project, self.target_dataset,
-                                                                 self.failed_tokenize_table)
+       | "Failed Tokenization" >> gh_bq.WriteFailedTokenizedData(self.failed_tokenize_table)
```
jlewi: Can we just do:

```
  beam.io.WriteToBigQuery(table=self.failed_tokenize_table,
                               schema=failed_tokenize_schema,
                               create_disposition=beam.io.BigQueryDisposition.CREATE_IF_NEEDED,
                               write_disposition=beam.io.BigQueryDisposition.WRITE_EMPTY)
```

We can replace WriteFailedTokenizedData
https://github.com/kubeflow/examples/blob/fa1311833c388b172093a82a1d8e4628c73b4fb2/code_search/src/code_search/dataflow/transforms/github_bigquery.py#L77

With an instance of [BigQuerySchema](https://github.com/kubeflow/examples/blob/fa1311833c388b172093a82a1d8e4628c73b4fb2/code_search/src/code_search/dataflow/transforms/bigquery.py#L45) 

This is a helper class for constructing a schema from a list of columns.

I think this is an improvement over the original code and I think its worth doing that rather than updating the underlying transforms.

code_search/src/code_search/dataflow/transforms/github_dataset.py
```diff
@@ -59,13 +54,11 @@ def expand(self, input_or_inputs):
...
 
-    if self.target_dataset:
-      logging.info("Writing results to BigQuery %s:%s.%s",
-                   self.project, self.target_dataset, self.pairs_table)
+    if self.pairs_table:
+      logging.info("Writing results to BigQuery %s", self.pairs_table)
       (flat_rows  # pylint: disable=expression-not-assigned
-        | "Save Tokens" >> gh_bq.WriteTokenizedData(self.project, self.target_dataset,
-                                                    self.pairs_table)
+        | "Save Tokens" >> gh_bq.WriteTokenizedData(self.pairs_table)
```
jlewi: Same comment as above about defining a schema and using WriteToBigQuery directly.

IronPan: done

code_search/src/code_search/dataflow/transforms/github_dataset.py
IronPan: done

jlewi:
Thanks
/lgtm
/approve

PR #407, opened 2018-12-06
remove obsolete PS in GPU jsonnet

Svendegroote91:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/407)

### DIFF
--- a/financial_time_series/tensorflow_model/GPU/train.jsonnet
+++ b/financial_time_series/tensorflow_model/GPU/train.jsonnet
@@ -44,22 +44,6 @@ local tfjob = {
           },
         },
       },
-      Ps: {
-        template: {
-          spec: {
-            containers: [
-              {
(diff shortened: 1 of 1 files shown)


jlewi:
Why is the PS no longer needed?

Svendegroote91:
Hi Jeremy,

In the example I am running a single training job (TFjob) to train a model and I want to showcase the difference in training time when running it on a GPU instead of CPU. The standard PS component is obsolete for what I want to showcase and therefore I would like to remove it (to keep it as simple as possible).

Let me know in case I got this wrong.

Best regards,
Sven

Op 12 dec. 2018 om 19:01 heeft Jeremy Lewi <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


Why is the PS no longer needed?

Issue #408, opened 2018-12-06
[object detection] Serve the model using TF-Serving cannot work


hougangliu:
When user build a new object detection image by following submit_job.md, everything runs well. But when serve the model using TF-Serving by following tf_serving_cpu.md, the server failed with error message _"Loading servable: {name: pets-model version: 1} failed: Not found: Op type not registered 'NonMaxSuppressionV3' in binary running on pets-model-v1-7c65cb44c-pg7wk. Make sure the Op and Kernel are registered in the binary running in this process."_

PR #409, opened 2018-12-06
reconcile tensorflow serving version

hougangliu:
Fixes: #408

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/409)

### DIFF
--- a/object_detection/docker/Dockerfile.training
+++ b/object_detection/docker/Dockerfile.training
@@ -43,7 +43,7 @@ RUN curl -O https://bootstrap.pypa.io/get-pip.py && \
     rm get-pip.py
 
 RUN pip --no-cache-dir install \
-        tensorflow
+        tensorflow==1.10.0
 
 RUN pip --no-cache-dir install \
         Cython \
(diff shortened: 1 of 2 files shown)


object_detection/tf_serving_cpu.md
```diff
@@ -17,12 +17,14 @@ MODEL_COMPONENT=pets-model
 MODEL_PATH=/mnt/exported_graphs/saved_model
 MODEL_STORAGE_TYPE=nfs
 NFS_PVC_NAME=pets-pvc
+TF_SERVING_VERSION=`kubectl -n kubeflow exec -ti tf-training-job-master-0 -- python -c 'from __future__ import print_function;import tensorflow;print(tensorflow.__version__, end="")'`
```
jlewi: Is there a cleaner way to handle this? Why do we wait until runtime to get the TF version? Shouldn't this be determined by the code that gets checked in?
https://github.com/kubeflow/examples/blob/master/object_detection/docker/Dockerfile.training#L45

So can't we just have the code pin consistent version of TF across the example?

hougangliu: Thanks for your comment. Agree with you. Since default OBJ_DETECTION_IMAGE tensorflow version is 1.10.0, we pin consistent version 1.10.0 of TF across the example.
Updated.

PR #418, opened 2018-12-08
fix bq table dupliation

IronPan:
Without the fix, table name was determined at compile time and would be same for all recurrent runs. The dataflow job will fail with table already exist error.

Since pipeline doesn't support inject unique camel case workflow name yet, manually replace '-' with '_' as a workaround. 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/418)

### DIFF
--- a/code_search/docker/ks/submit_code_embeddings_job.sh
+++ b/code_search/docker/ks/submit_code_embeddings_job.sh
@@ -41,12 +41,9 @@ names=(cluster dataDir functionEmbeddingsBQTable functionEmbeddingsDir modelDir 
 source "${DIR}/parse_arguments.sh"
 source "${DIR}/initialize_kubectl.sh"
 
-# BQ doesn't support dash. Replace with underscore
-functionEmbeddingsBQTable=$(echo "${functionEmbeddingsBQTable}" | tr - _)
-
 # Apply parameters
 ks param set ${component} dataDir ${dataDir} --env ${ksEnvName}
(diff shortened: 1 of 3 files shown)


code_search/docker/ks/submit_code_embeddings_job.sh
```diff
@@ -41,9 +41,12 @@ names=(cluster dataDir functionEmbeddingsBQTable functionEmbeddingsDir modelDir
 source "${DIR}/parse_arguments.sh"
 source "${DIR}/initialize_kubectl.sh"
 
+# BQ doesn't support dash. Replace with underscore
+functionEmbeddingsBQTable=$(echo "${functionEmbeddingsBQTable}" | tr - _)
+
 # Apply parameters
 ks param set ${component} dataDir ${dataDir} --env ${ksEnvName}
-ks param set ${component} functionEmbeddingsBQTable ${functionEmbeddingsBQTable} --env ${ksEnvName}
+ks param set ${component} functionEmbeddingsBQTable ${project}:${functionEmbeddingsBQTable} --env ${ksEnvName}
```
jlewi: Shouldn't project be part of the functionEmbeddingBQTable name? i.e. shouldn't functionEmbeddingsBQTable be the full table spec?

IronPan: yeah i agree it should be. however the line above

> functionEmbeddingsBQTable=$(echo "${functionEmbeddingsBQTable}" | tr - _)

will replace the project from code-search-demo to code_search_demo. so i separate them to workaround the issue. 

I revert the change since I find some other parameter placeholder i can use for specifying the BQ table name.

PR #423, opened 2018-12-09
Update server DNS for ks web app pipeline environment

IronPan:
Without the change, I got an error when configuring argocd: 
InvalidSpecError: cluster 'https://kubernetes.default' has not been configured

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/423)

### DIFF
--- a/code_search/ks-web-app/app.yaml
+++ b/code_search/ks-web-app/app.yaml
@@ -8,7 +8,7 @@ environments:
     path: cs_demo
   pipeline:
     destination:
-      namespace: cs-web-app
+      namespace: kubeflow
       server: https://35.185.115.154
     k8sVersion: v1.10.9
     path: pipeline
--- a/code_search/ks-web-app/app.yaml
+++ b/code_search/ks-web-app/app.yaml
@@ -9,7 +9,7 @@ environments:
   pipeline:
     destination:
       namespace: cs-web-app
-      server: https://kubernetes.default
+      server: https://35.185.115.154
     k8sVersion: v1.10.9
     path: pipeline
 kind: ksonnet.io/app


jlewi:
Where are you using this environment? I thought the purpose of the pipeline environment was to be used from within the cluster which is why it was using the internal name "kubernetes" which will only resolve if you are running on cluster.

If we switch to using the external IP address of the master; do we even need the pipeline environment or could we just reuse the existing cs_demo enviornment?

If there are parameter values specific to the pipeline environment then I think it makes sense to keep it as a separate environment.

IronPan:
I thought it would be nice to separate them out so we can still have a stable cs_demo web UI while having a pipeline env to accept the constant changes like
https://github.com/kubeflow/examples/pull/426

jlewi:
Do you need to change the namespace then?

Right now this is pointing at the namespace cs-web-app so you would end up messing with the same instance as the stable instance.

IronPan:
make sense. switch to kubeflow ns

jlewi:
Sorry I should have been more clear.

I think the web app should be deployed in a different namespace from kubeflow. This is just better isolation and cleaner.

Right now we are using cs-web-app as the namespace for the web app.

We already have the cs_demo environment pointing at that namespace. So if we are just using that namespace then we don't need a separate environment for the Kubeflow pipeline.

if we use a separate environment then we should probably point it at a different namespace that's not cs-web-app and not Kubeflow.

I think my preference would actually be to just have the pipeline update the app in the namespace cs-web-app and not have two different instances of the app running or use a different ksonnet environment.

I'm not too worried about stability. So if we can cut down the amount of complexity/environments I think that's good.

So my suggestion would be to just delete the pipeline environment and update the kubeflow pipeline to use the cs_demo environment.

WDYT?

jlewi:
Lets

PR #427, opened 2018-12-10
Minor fixes to the notebook.

jlewi:
* Need to fix the import and compile commands.

* Check if an experiment with the name already exists.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/427)


jlewi:
I see looks like there is a conflict now.

PR #432, opened 2018-12-11
Fix the web app.

jlewi:
* We need to set the parameters for the model and index.

  * It looks like when we split up the web app into its own ksonnet app
    we forgot to set the parameters.

* SInce the web app is being deployed in a separate namespace we need to
  copy the GCP credential to that namespace. Add instructions to the
  demo README.md on how to do that.

* It looks like the pods were never getting started because the secret
  couldn't be mounted.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/432)

### DIFF
--- a/code_search/demo/README.md
+++ b/code_search/demo/README.md
@@ -26,6 +26,16 @@ We are using the following project
    ```
    ks12 apply  cs_demo -c search-index-server
    ```
+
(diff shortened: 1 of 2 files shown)


IronPan:
/lgtm
although it would be easier if the web app is deployed in kubeflow namespace?

PR #439, opened 2018-12-17
create pv for pets-pv

hougangliu:
For a lot of user k8s clusters, dynamic volume provisioning isn't
enabled. So the newcomer may be blocked since pets-pv will keep
Pending. We can guide them to create a nfs PV as an option.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/439)


jlewi:
Thanks for submitting a fix! 

Are you assuming the user already has an NFS cluster? 

I think the example is assuming the user has a default storage class defined which will automatically provision a PV given a PVC.

Can we provide users instructions for how to test whether they have a default storage class defined
e.g
kubectl get storageclass -o yaml standard

Then rather than assuming a user has NFS can we just point them at Kubernetes docs telling them to create a PV or PVC?
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volumes

gyliu513:
@jlewi Yes, we can add some detail for end user to check if they have some default storageclass, but usually, a test cluster do not have storageclass as it may request some integration to external storage, so using NFS is a very simple way for tutorial.

So how about the following:
1) Add some detail for how to check storageclass in the cluster.
2) Still keep @hougangliu 's fix for NFS which act as another option for this tutorial?

hougangliu:
Generally nfs server is easy to set up even there is no an available one. And maybe NFS PV is the most easy method to create a PV, even we add link to tell them how to create a PV.
Also we always don't require user do extra effort about k8s operation. The user expects run this demo just by copy & paste.

Updated to tell user how to check if a default storageclass is defined

object_detection/setup.md
```diff
@@ -66,6 +66,35 @@ does not support this feature you can modify the `accessMode` value to create th
...
+
+Otherwise you can find that the PVC remains `Pending` status.
+
+```
+$ kubectl get pvc pets-pvc -n kubeflow
+NAME        STATUS    VOLUME    CAPACITY   ACCESS MODES   STORAGECLASS   AGE
+pets-pvc    Pending                                                      28s
+```
+
+You can create a PV with your NFS server to make the PVC work.
```
gyliu513: s/You can create a PV with your NFS server to make the PVC work./If your cluster do not have `storageclass`, you can create a PV with your NFS server to make the PVC work./

hougangliu: Thanks! Updated

jlewi:
@hougangliu @gyliu513 Can you explain the situation you are targeting where you think NFS is easy to setup? e.g. why are you assuming that users have an NFS setup or that its easy to setup?

In Cloud setting up NFS is non-trivial; and if you're using a newly created project I wouldn't expect the user to have access to NFS. I believe all clouds have some form of persistent disk which is far easier to use.

So that leaves OnPrem clusters and local installs (e.g. minikube).

Is it reasonable to assume on prem clusters have NFS?  If an on prem deployment doesn't have NFS is it reasonable to expect a user can set it up?

Since we don't know what a user's setup is, why don't we just point users at the most relevant instructions for figuring out what might work in their particular use case as opposed to assuming they have NFS or can set it up?

gyliu513:
Thanks @jlewi 

We are now testing in on-prem clusters, setting up with minikube or some other kubernetes distributions, and for an on-prem cluster, it is pretty easy to set up an NFS cluster.

Here the document is just setting NFS as an optional, they can use NFS if they do not have a storageclass, comments?

hougangliu:
@jlewi updated

jlewi:
Thanks!
/lgtm
/approve

PR #445, opened 2018-12-26
Update pytorch_mnist example to use v1beta1

gabrielwen:
Currently if we deploy kubeflow from HEAD, tf operator is default at v1beta1.  Updating example here to match with it.


texasmichelle:
Is there any reason we need to keep the old v1alpha2 files around? If not, let's remove them.

gabrielwen:
> Is there any reason we need to keep the old v1alpha2 files around? If not, let's remove them.

removed :)

gabrielwen:
This PR is ready.

Issue #295, opened 2018-11-02
[code_search] Run E2E test as part of CI


jlewi:
#291 added similarity_transformer_test.py which runs through data preprocessing, training, and model export.

We should integrate this into the CI system.

The main things we need to do are

1. Create a Docker image with the suitable dependencies (e.g TensorFlow in it)
    We should probably base this on the image used to run the code.

1. We can modify test_py_checks to allow invoking specific tests. Then we can easily add the test to our E2E workflow. ( see also kubeflow/testing#240)

jlewi:
#446 Will run the python test but the python test isn't emitting a junit file to report the test results to Prow.

For that I think we want to create a test runner for prow as discussed in kubeflow/testing#270

PR #446, opened 2018-12-27
An Argo workflow to use as the E2E test for code_search example.

jlewi:
An Argo workflow to use as the E2E test for code_search example.
    
* The workflow builds the Docker images and then runs the python test
   to train and export a model
    
* Move common utilities into util.libsonnet.
    
* Add the workflow to the set of triggered workflows.
    
* Update the test environment used by the test ksonnet app; we've since
      changed the location of the app.
    
Related to #295
    
* Refactor the jsonnet file defining the GCB build workflow
    
   * Use an external variable to conditionally pull and use a previous
        Docker image as a cache
    
   * Reduce code duplication by building a shared template for all the different
        workflows.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/446)


jlewi:
kubeflow/testing#265 is merged; lets try again.

/test all

jlewi:
This is ready for review.

/hold cancel
/assign @texasmichelle

Issue #92, opened 2018-04-24
[GH Issue] E2E test


jlewi:
We should add E2E testing for the GH issue example to make sure it doesn't break.

Some things to test
 
 - Make sure we can deploy all the components using ksonnet
 - Make sure training runs (just run a couple steps)
 - Test predict RPC generate responses

jlewi:
So I think a good place to start would be testing the things the tutorial tells users todo
https://github.com/kubeflow/examples/tree/master/github_issue_summarization

I think we can start by 

1. Building a docker image
1. Submitting a TFJob to train the model for a small number of steps
1. Verifying the TFJob completes successfully

Rather than have the E2E test spin up a Kubeflow cluster; I think we should have a set of Kubeflow clusters corresponding to different Kubeflow versions in our ci project which we then reuse for E2E example testing.

jlewi:
We have a test for training here that we can run.
https://github.com/kubeflow/examples/blob/master/github_issue_summarization/notebooks/train_test.py

PR #449, opened 2018-12-29
Setup continuous building of Docker images for GH Issue Summarization Example

jlewi:
Setup continuous building of Docker images and testing  for GH Issue Summarization Example.
    
* This is the first step in setting up a continuously running CI test.

* Add support for building the Docker images using GCB; we will use GCB
  to trigger the builds from our CI system.

  * Make the Makefile top level (at root of GIS example) so that we can
    easily access all the different resources.

* Add a .gitignore file to avoid checking in the build directory used by
  the Makefile.

* Define an Argo workflow to use as the E2E test.

Related to #92: E2E test & CI for github issue summarization

* Trigger the test on pre & post submit

* Dockerfile.estimator don't install the data_download.sh script
  * It doesn't look like we are currently using data_download.sh in the
    DockerImage
  * It looks like it only gets used vias the ksonnet job which mounts the
    script via a config map

  * Copying data_download.sh to the Docker image is currently weird
    given the organization of the Dockerfile and context.

* Copy the test_data to the Docker images so that we can run the test
  inside the images.

* Invoke the python unittest for training from our CI system.

  * In a follow on PR we will update the test to emit a JUnit XML file to
    report results to prow.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/449)


jlewi:
/assign @texasmichelle 

This is ready for review.

jlewi:
/assign @zhenghuiwang 
@zhenghuiwang could take another look and re-lgtm? I had to fix the test.

PR #468, opened 2019-01-10
Fix the model training of ames-housing example

zhenghuiwang:
Fix two issues:
1.The training data is not available in either the image nor the PV.
- Fix it by adding the data set to the docker file. (the data set is small so save the effort to add step to download it to PV and access it via PV.)
- Build a new docker image and make it public.

2. Existing 'py-pod` fails to write model file to PV.
- `py-pod` keeps failing because the model training is a one-time work. The right workload type should be `Job`.
- Add `py-job` to replace `py-pod`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/468)


xgboost_ames_housing/README.md
```diff
@@ -35,7 +35,7 @@ directly download the image.
 
 ```
 IMAGE_NAME=ames-housing
-VERSION=v1
+VERSION=latest
```
richardsliu: Probably better to pin a specific version instead of using latest.

xgboost_ames_housing/Dockerfile
```diff
@@ -21,6 +21,9 @@ FROM python:2.7
 RUN apt-get update
 RUN apt-get install -y git make g++
 
+# Download the training set to local directory 'ames_dataset'.
+COPY ames_dataset/ ./ames_dataset
```
richardsliu: Do we also need to copy the python code to the container? Seems kind of weird that we need to download the actual python script from Github.

xgboost_ames_housing/README.md
zhenghuiwang: In case of codelab/examples instructions, it is always better to use latest tag instead of a specific tag, because readers should not care about the version of image.  You can see that [hello_world example](https://docs.docker.com/get-started/) for docker.

Internally, we can add various version tags to the images if we want to track the image changes, as one image can have multiple tags.

xgboost_ames_housing/Dockerfile
zhenghuiwang: Good point. Change it to copy the code needed only from local files.

Issue #460, opened 2019-01-07
[mnist] E2E tests


jlewi:
We need to have E2E tests to verify the mnist example works. Tests should ensure

1. Training works
1. Deploying model works
1. Sending predictions works

This is P1 because we see lots of customer demand for xgboost so we need to know that its working against the latest Kubeflow release.

I'm working on adding tests to the GitHub issue summarization example. We can use that as a model for adding tests to this example.

In addition here are some other resources:

* [General instructions](https://github.com/kubeflow/testing#writing-an-argo-workflow-for-an-e2e-test) for writing an Argo workflow to be used as an E2E test

* [Guide To Writing Tests for E2E examples](http://bit.ly/kf_example_testing_guide) - This is in an in progress doc to come up with instructions for writing E2E tests.

I'm bumping this to priority P1 because we currently use mnist as part of our getting started guides; e.g https://www.kubeflow.org/docs/guides/gke/gcp-e2e/

jlewi:
It looks like some refactoring of the mnist example might be needed as well.

* There is currently an Argo workflow to do training
  https://github.com/kubeflow/examples/blob/master/mnist/model-train.yaml

There doesn't appear to be a manifest to just train the model.

For comparison our PyTorch example has separate manifests for CPU/GPU but these are YAML not ksonnet

https://github.com/kubeflow/examples/tree/master/pytorch_mnist/training/ddp/mnist/cpu/v1alpha2

The E2E guide for GKE has the user manually create and edit a jsonnet file.
https://www.kubeflow.org/docs/guides/gke/gcp-e2e/#train-the-model-on-kubernetes-engine

We could just use that.

jlewi:
Tests are committed but they aren't showing up in the testgrid
https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-examples-postsubmit

We should enable periodic tests for the examples and verify the tests are running regularly.

jlewi:
Tests are in the periodic dashboard.
Dashboard is setup
https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-periodic-examples

Tests are running.
https://gubernator.k8s.io/build/kubernetes-jenkins/logs/kubeflow-periodic-examples/1

PR #469, opened 2019-01-11
Update serving in mnist example; use 0.4 and add testing.

jlewi:
* Add the TFServing component
* Create TFServing components.

* The model.py code doesn't appear to be exporting a model in saved model
  format; it was a missing a call to export.

  * I'm not sure how this ever worked.

* It also looks like there is a bug in the code in that its using the cnn input fn even if the model is the linear one. I'm going to leave that as is for now.

* Create a namespace for each test run; delete the namespace on teardown
* We need to copy the GCP service account key to the new namespace.
* Add a shell script to do that.

Related to #460

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/469)


jlewi:
/assign @zhenghuiwang 
This is ready for review.

mnist/ks_app/app.yaml
```diff
@@ -2,14 +2,28 @@ apiVersion: 0.3.0
 environments:
   jlewi:
     destination:
-      namespace: kubeflow
+      namespace: jlewi
```
zhenghuiwang: Could we use a more meaningful namespace & path, such as `kf-mnist'?

mnist/ks_app/vendor/kubeflow/tf-serving@fed535eaa276220e4edf59530c0629f4375a40a9/README.md
```diff
@@ -0,0 +1,73 @@
...
+* [Using Prototypes](#using-prototypes)
+  * [io.ksonnet.pkg.tf-serving](#io.ksonnet.pkg.tf-serving)
+
+## Quickstart
+
+*The following commands use the `io.ksonnet.pkg.tf-serving` prototype to generate Kubernetes YAML for tf-serving, and then deploys it to your Kubernetes cluster.*
+
+First, create a cluster and install the ksonnet CLI (see root-level [README.md](rootReadme)).
+
+If you haven't yet created a [ksonnet application](linkToSomewhere), do so using `ks init <app-name>`.
```
zhenghuiwang: Add the link?

test/copy_secret.sh
```diff
@@ -0,0 +1,33 @@
...
+set -e
+SOURCE=$1
+DEST=$2
+NAME=$3
+
+usage() {
+	echo copy_secret "<source namepspace> <dest namespace> <secret name>"
+}
+
+if [ -z ${SOURCE} ]; then
```
zhenghuiwang: how about using `||` to combine these 3 checking statements.

PR #470, opened 2019-01-12
Add richard and zhenghui to approvers for kubeflow/examples

zhenghuiwang:
Add richard and zhenghui to approvers for kubeflow/examples

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/470)

### DIFF
--- a/xgboost_ames_housing/OWNERS
+++ b/xgboost_ames_housing/OWNERS
@@ -1,9 +1,7 @@
 approvers:
-  - gaocegege
   - jlewi
-  - lluunn
-  - texasmichelle
   - richardsliu
   - zhenghuiwang
 reviewers:
-  - cwbeitel
+  - richardsliu
+  - zhenghuiwang
--- a/xgboost_ames_housing/OWNERS
+++ b/xgboost_ames_housing/OWNERS
@@ -1,0 +1,9 @@
+approvers:
+  - gaocegege
+  - jlewi
+  - lluunn
+  - texasmichelle
+  - richardsliu
+  - zhenghuiwang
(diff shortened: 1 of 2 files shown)


jlewi:
I would suggest adding an OWNERs file in the xgboost example directory.
Having smaller more targeted OWNERs files helps auto-assign reviews to reviewers familiar with the code.

zhenghuiwang:
good suggestion. add a new OWNER file.

xgboost_ames_housing/OWNERS
```diff
@@ -0,0 +1,9 @@
+approvers:
+  - gaocegege
```
jlewi: Why gaocegege, jlewi, lunkai?

I would suggest just listing you and richard for now since you are the only two people actively working on this example.

We can others in the future if needed be.

zhenghuiwang: makes sense. changed it.

xgboost_ames_housing/OWNERS
```diff
@@ -0,0 +1,7 @@
+approvers:
+  - jlewi
+  - richardsliu
+  - zhenghuiwang
+reviewers:
```
jlewi: nit: You don't need to list me here; I'm already in the parent file and I don't have any special familiarity with the xgboost example.

don't list approvers as reviewers; all approvers are automatically considered reviewers.

zhenghuiwang: I should have read the Owner guideline more carefully. My fault... it should be fine this time.

Issue #455, opened 2018-12-29
[GH Issue] E2E test for deploying the model and sending predictions


jlewi:
Related to #92: E2E test for GitHub issue Summarization

Our E2E test should deploy the model and send some predictions to verify it works.

One complication is figuring out where to run the client and server so they can communicate. If we don't run them within the same cluster then we run into complications with getting access to the server inside the cluster.

Here are some potential solutions.

1. If we deploy the model on of our test clusters (e.g. https://kf-v0-4-n00.endpoints.kubeflow-ci.cloud.goog) which has IAP enabled then we can send predictions via IAP and rely on Ambassador to route the predictions to the model.

1. We can deploy the model in the same namespace as where the test runner is running so the test runner can directly send predictions.

1. The test runner can deploy the model in the cluster and can run a second job in that same cluster to send predictions to the model.

jlewi:
Seldon provides some helpful scripts for testing.
https://github.com/SeldonIO/seldon-core/blob/master/docs/api-testing.md

Maybe we can use those?

PR #472, opened 2019-01-12
GitHub Summarization Seldon Update

cliveseldon:
* Changed to use s2i for building image and updated the Makefle
 * Updated Keras model to ensure it works in latest Flask by calling `_make_predict_function()` see keras issue discussion [here](https://github.com/keras-team/keras/issues/6462)
 * Updated docs

Related to #455 E2E test for deploying models

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/472)


jlewi:
@cliveseldon If this is ready for review can you remove WIP from the title please?

github_issue_summarization/03_serving_the_model.md
```diff
@@ -14,41 +14,59 @@ You have two options for getting a model server
...
 
-1. You can build your own model server as discussed below
+1. You can build your own model server as discussed below. For this you will need to install the [Source2Image executable s2i](https://github.com/openshift/source-to-image).
 
 
-## Wrap the model into a seldon-core microservice
+## Wrap the model into a Seldon Core microservice
 
-cd into the notebooks directory and run the following docker command. This will create a build/ directory.
+Set a couple of environment variables tospecify the GCP Project and the TAG you want to build the image for:
```
jlewi: nit: "tospecify" -> "to specify"

github_issue_summarization/03_serving_the_model.md
```diff
@@ -83,12 +101,25 @@ Now that you have seldon core deployed, you can deploy the model using the
 ```bash
 ks generate seldon-serve-simple-v1alpha2 issue-summarization-model \
```
jlewi: Should we just check in the resulting ksonnet module into the ksonnet app
https://github.com/kubeflow/examples/tree/master/github_issue_summarization/ks_app

And then tell users to just add an environment and then deploy the component?

(We can also address that in a follow on PR).

github_issue_summarization/notebooks/Makefile
```diff
@@ -70,38 +70,39 @@ build:
...
-	docker run -v $(BUILD_DIR):/my_model seldonio/core-python-wrapper:0.7 /my_model IssueSummarization $(TAG) gcr.io --base-image=python:3.6 --image-name=$(MODEL_IMG_NAME)
-	# We don't use the script generated by Seldon because that script won't get updated by make to reflect the change
-	# in the desired tag.
-	docker build -t $(MODEL_IMG):$(TAG) -f $(BUILD_DIR)/build/Dockerfile $(BUILD_DIR)/build
+build-model-image: download-model $(BUILD_DIR)/seq2seq_utils.py $(BUILD_DIR)/IssueSummarization.py $(BUILD_DIR)/requirements.txt $(BUILD_DIR)/environment_seldon_rest
+	cd $(BUILD_DIR) && s2i build -E environment_seldon_rest . seldonio/seldon-core-s2i-python36:0.4 $(MODEL_IMG):$(TAG)
 	@echo built $(MODEL_IMG):$(TAG)
 
-push-model-image: build-model-image
+start-docker-model-image:
```
jlewi: no action needed; just an FYI

My current thinking is that a better way to begin testing is by writing a manual test; e.g. a python file that can send predictions and validate the results; something like
https://github.com/kubeflow/examples/blob/master/mnist/testing/deploy_test.py

To begin with this can be run as a manual test where the model can be running either locally (via docker) or on a cluster. We can use kubectl port-forward to setup a local port if needed to use.

jlewi:
Looks like some lint issues

cliveseldon:
The lint issue are to do with my addition of lines like `self.decoder_model._make_predict_function()`. See the link to the Keras issue in the PR description. If we can't bypass the lint checks then the only solution I think is to add a dummy prediction after the model is loaded I think.

jlewi:
@cliveseldon You should be able to disable the lint checks on those lines by just adding an inline comment e.g.

```
 self.seq2seq_model._make_predict_function()  # pylint: disable=protected-access
```

PR #476, opened 2019-01-14
Fix xgboost example for hyperparameter tuning

richardsliu:
- Don't shuffle the input data to reduce randomness between trials
- Output mean_absolute_error in the format that katib metric collector expects

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/476)


xgboost_ames_housing/housing.py
```diff
@@ -61,7 +62,7 @@ def train_model(train_X,
 def eval_model(model, test_X, test_y):
   """Evaluate the model performance."""
   predictions = model.predict(test_X)
-  print("MAE on test: {:.2f}".format(mean_absolute_error(predictions, test_y)))
+  print("mean_absolute_error={:.2f}".format(mean_absolute_error(predictions, test_y)))
```
jlewi: Should we use the logging module rather than print?

richardsliu: Fixed

PR #478, opened 2019-01-15
Fix model serving part of xgboost_ams_housing example.

zhenghuiwang:
Issues fixed:

- Update the seldon serving image build command with `s2i` and its environment file.
- Fix `ks generate` command.
- Remove hardcoded image project name.


zhenghuiwang:
/assign @richardsliu 
/assign @jlewi 

It is ready for your review.

PR #479, opened 2019-01-15
Create an E2E test for TFServing using the rest API

jlewi:
* We use the pytest framework because
  1. it has really good support for using command line arguments
  2. can emit junit xml file to report results to prow.

Related to #270: Create a generic test runner

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/479)

### DIFF
--- a/mnist/testing/predict_test.py
+++ b/mnist/testing/predict_test.py
@@ -22,10 +22,10 @@ Manually running the test
 import json
 import logging
 import os
+import subprocess
 import requests
 from retrying import retry
 import six
-import subprocess
 
 from kubernetes.config import kube_config
 from kubernetes import client as k8s_client
--- a/test/workflows/components/mnist.jsonnet
+++ b/test/workflows/components/mnist.jsonnet
@@ -33,8 +33,7 @@ local defaultParams = {
   // But if we leave namespaces up; then we end up leaving the servers up which
   // uses up CPU.
   //
-  // DO NOT SUBMIT set back to true
-  deleteNamespace: false,
+  deleteNamespace: true,
 };
 
 local params = defaultParams + overrides;
@@ -483,6 +482,8 @@ local workflow = {
   },
   spec: {
     entrypoint: "e2e",
+    // Have argo garbage collect old workflows otherwise we overload the API server.
+    ttlSecondsAfterFinished: 7 * 24 * 60 * 60,
     volumes: [
       {
         name: "github-token",


mnist/testing/predict_test.py
```diff
@@ -0,0 +1,91 @@
...
+  # We proxy the request through the APIServer so that we can connect
+  # from outside the cluster.
+  url = ("https://{master}/api/v1/namespaces/{namespace}/services/{service}:8500"
+         "/proxy/v1/models/mnist:predict").format(
+           master=master, namespace=namespace, service=service)
+  r = requests.post(url, json=instances, headers=headers, verify=False)
+
+  if r.status_code == requests.codes.NOT_FOUND:
+    logging.info("Request to %s returned 404", url)
+    return
```
zhenghuiwang: Do we want to raise an error instead of `return`? If we return here, will argo thinks this file execution is successful?

test/workflows/components/mnist.jsonnet
```diff
@@ -15,7 +15,7 @@ local defaultParams = {
   dataVolume: "kubeflow-test-volume",
 
   // Default step image:
-  stepImage: "gcr.io/kubeflow-ci/test-worker:v20190104-f2a1cdf-e3b0c4",
+  stepImage: "gcr.io/kubeflow-ci/test-worker/test-worker:v20190116-b7abb8d-e3b0c4",
```
zhenghuiwang: Curious: how is worker image being built? Is it automatically built or manually?

mnist/testing/predict_test.py
```diff
@@ -0,0 +1,91 @@
...
+
+from kubernetes.config import kube_config
+from kubernetes import client as k8s_client
+
+import pytest
+
+from kubeflow.testing import util
+
+def test_predict(master, namespace):
+  app_credentials = os.getenv("GOOGLE_APPLICATION_CREDENTIALS")
```
zhenghuiwang: Is it better to encapsulate line 36-59 as a function in `kubeflow.testing.util` for reuse?

mnist/testing/predict_test.py
jlewi: Good catch thanks

test/workflows/components/mnist.jsonnet
jlewi: Right now its all manual. We could build it on successful post submit like we do our examples we just need to add support for that.

mnist/testing/predict_test.py
jlewi: Once the code stabilizes probably. The problem right now is that putting it in kubeflow.testing means splitting it across two different repos and thus two different PRs. This is pretty painful during the early stages when we detect bugs and need to fix them.

e.g some of the PY3 issues only cropped up on the test system because I use PY2.7 locally. Similarly, the gcloud service account code is specific to running on the test cluster rather than locally.

zhenghuiwang:
This pr looks good, except some linting errors.

jlewi:
@zhenghuiwang Finally got the tests to pass. PTAL.

PR #484, opened 2019-01-19
Build images for xgboost-ames-housing

zhenghuiwang:
This PR adds configs and makefile for building a seldon serving image via GCB, which will be part of the e2e test for the housing xgboost example.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/484)


zhenghuiwang:
Tested it locally.

/assign @jlewi 
/assign @r2d4

PR #507, opened 2019-02-13
import of Pipelines Github issue summarization examples & tutorial

amygdala:
@texasmichelle 
I still need to double check that I got all the necessary bits and pieces imported before this is ready to pull in. (I need to do some pylinting as well-- will do that tomorrow.)   
But wanted to create the PR so that you can see the structure of the new stuff.  Under the new `pipelines` subdir there are two dirs, `components` and `example_pipelines`.  `components` contains the image source, Dockerfiles, and build scripts for each pipeline step.

Note that the PR includes a .md version of the codelab with Cloud Shell markup (`tutorial.md`). (The Cloud Shell markup isn't rendered by Github). The README includes a link for how to launch the tutorial in the cloud console.

Update: also, reminder to us that the links in g.co/codelabs/kubecon18 should be updated once this is in. I'm keeping the code in its current repo for now, so nothing will break.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/507)


DanSanche:
I ran through this codelab the other night in the process of creating the [MNIST pipeline example](https://github.com/DanSanche/examples/tree/mnist-pipelines). I had a couple issues, so this might be a good place to discuss it:

- The serve/webapp components need permissions to run 'kubectl apply' within the container. How is this permission being handled? I don't see anything in the pipeline script about it, and I had some errors when I got to this step in the codelab
- is there a reason you wrote the tf-serving component yourself, when it seems like the Pipelines examples already has a [container image that does the same thing?](https://github.com/kubeflow/pipelines/blob/603597d65b1092d28d77111051e1175c40d77fea/samples/tfx/taxi-cab-classification-pipeline.py#L133) (related: do you know if pipelines has plans to make launching kubernetes resources part of the dsl? This seems like a major gap if long-running serving is meant to be a supported pipelines use case)
- Is there a way to make this work without GPUs enabled? Maybe even have a conditional inside the pipeline that checks for GPU resources? I had a lot of trouble finding a region that supports GPUs and getting everything working, and I imagine it would be much harder (and more expensive!) to set up GPUs outside Google

amygdala:
Just a meta-note that I will return soon to fix the remaining pylint issues etc.  (and reply to Dan's questions).  
@DanSanche , how did you create the cluster in which you saw the 'kubectl apply' error? I know what you're talking about in the general case (will add more a bit later), but, for me the permissions were okay when running that command via a GKE launcher-created cluster.

DanSanche:
I believe I used Click-to-Deploy, but it was a while ago now. If you want I can try to reproduce again next week, maybe I set something up wrong

amygdala:
(apologies for all the linting attempts -- I'm using the same .pylintrc as this repo, but for some reason my local runs don't flush out the same errors)

amygdala:
Apparently pylint gives different result depending upon what packages are installed in the environment, which explains a lot. I'm getting inconsistent linting results locally vs the presubmit tests, so I think am just going to try to suppress the ordering errors instead.

texasmichelle:
Digging into the `kubeflow-testing` cluster in the `kubeflow-ci` project, I'm seeing these errors:

```ERROR: (gcloud.container.clusters.get-credentials) ResponseError: code=404, message=Not found: projects/kubeflow-ci/zones/us-east1-d/clusters/kf-v0-4-n00.
No cluster named 'kf-v0-4-n00' in kubeflow-ci.
```

```
TypeError: preprocess_text() got an unexpected keyword argument 'transliterate'
```

```
UnboundLocalError: local variable 'transformed_data' referenced before assignment
```

texasmichelle:
2 unrelated issues:
* The [training_with_keras](https://github.com/kubeflow/examples/blob/master/github_issue_summarization/notebooks/train_test.py#L9) test is failing
* This cluster doesn't exist: `kf-v0-4-n00`. Do you know if that is created dynamically? Would updating it [here](https://github.com/kubeflow/examples/blob/master/test/workflows/components/gis.jsonnet#L23) resolve the issue?

texasmichelle:
I wonder if they're actually related. Try setting these values [here](https://github.com/kubeflow/examples/blob/master/test/workflows/components/gis.jsonnet#L22-L23):

```
  kfZone: "us-east1-b",
  kfCluster: "kf-vmaster-n00",
```

Issue #515, opened 2019-02-22
mnist training E2E test failing; chief replica fails


jlewi:
The training step for mnist is failing.
Here's a failed run from a recent periodic run
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-periodic-examples-mnist-90-70fd?tab=workflow&nodeId=kubeflow-periodic-examples-mnist-90-70fd-1605631456&sidePanel=logs%3Akubeflow-periodic-examples-mnist-90-70fd-1605631456%3Amain

The TFJob is in a failed status

```
   {
       "status": "True",
       "lastUpdateTime": "2019-02-22T10:27:54Z",
       "lastTransitionTime": "2019-02-22T10:27:54Z",
       "reason": "TFJobFailed",
       "message": "TFJob mnist-test-90 has failed because 1 Chief replica(s) failed.",
       "type": "Failed"
     }
   ],
```

Events indicate the chief pod was createad

```
'kind': None,
'last_timestamp': datetime.datetime(2019, 2, 22, 10, 27, 47, tzinfo=tzlocal()),
'message': 'Created pod: mnist-test-90-chief-0',
'metadata': {'annotations': None,
             'cluster_name': None,
             'creation_timestamp': datetime.datetime(2019, 2, 22, 10, 27, 48, tzinfo=tzlocal()),
             'deletion_grace_period_seconds': None,
             'deletion_timestamp': None,
             'finalizers': None,
             'generate_name': None,
             'generation': None,
             'initializers': None,
             'labels': None,
             'name': 'mnist-test-90.1585a8d24fb78cab',
             'namespace': 'mnist-90',
             'owner_references': None,
             'resource_version': '1397',
             'self_link': '/api/v1/namespaces/mnist-90/events/mnist-test-90.1585a8d24fb78cab',
             'uid': '8048d797-368c-11e9-932e-42010af0008a'},
'reason': 'SuccessfulCreatePod',
'related': None,
'reporting_component': '',
'reporting_instance': '',
'series': None,
'source': {'component': 'tf-operator', 'host': None},
'type': 'Normal'}
```
I couldn't find any logs for that pod though.

Event logs indicate pod exited with code 1 `'Pod: mnist-90.mnist-test-90-chief-0 exited with code 1',

jlewi:
The missing logs might be due to kubeflow/kubeflow#2300

[...]

PR #516, opened 2019-02-23
Trigger tests to print out logs of mnist trainning

jinchihe:
Due to the problem in the mnist training steps of kubeflow-examples-presubmit tests, and without logs. I just just send the PR to have a try, and trigger some tests to print out the related error logs. Thanks.

The PR updated ` mnist/ks_app/components/train.jsonnet` by adding a `the` in description to trigger the training tests. and updated `mnist/testing/tfjob_test.py` to printed error logs of the tfjobs related pod for debugging.

/do-not-merge

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/516)

### DIFF
--- a/mnist/ks_app/components/train.jsonnet
+++ b/mnist/ks_app/components/train.jsonnet
@@ -1,6 +1,6 @@
 // Component to train a model.
 //
-// Parameters are used to control training
+// Parameters are used to control the training
 //   image: Docker iamge to use
 //   modelDir: Location to write the model this can be a local path (e.g. to a PV)
 //             or it can be any filesystem URI that TF understands (e.g GCS, S3, HDFS)
--- a/mnist/testing/tfjob_test.py
+++ b/mnist/testing/tfjob_test.py
@@ -28,6 +28,7 @@ import logging
 import os
 
 from kubernetes import client as k8s_client
+from kubernetes.client import rest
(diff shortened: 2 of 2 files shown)


jlewi:
Do you want to add a TODO explaining why we need to explicitly fetch the logs; i.e kubeflow/kubeflow#2300 ?

And then merge this as a workaround?

jinchihe:
> Do you want to add a TODO explaining why we need to explicitly fetch the logs; i.e [kubeflow/kubeflow#2300](https://github.com/kubeflow/kubeflow/issues/2300) ?
> 
> And then merge this as a workaround?

Hello @jlewi Thanks for your comments.

But I think the PR has nothing with the kubeflow/kubeflow#2300. This is not a work around. while #515 investigation, I found the logs of `kubeflow-examples-presubmit-mnist-xxx.tfjob-test` can be got even if without 2 minutes delay. But no relevant pod logs from argo ui, you know, if the tfjobs status is failed, that's very useful to check the relevant pod logs to get root course if tfjobs failed, such as if there is a error during the trainning in the training pod...

The purpose of the PR changes is print out the logs of the related pods only if the tfjobs failed, so that user can read logs from argo UI and debug easily.
 
BTW, The PR also updated ` mnist/ks_app/components/train.jsonnet` by adding a `the` in description to trigger the training tests, in fact I think that should be have. :-)

PR #519, opened 2019-02-27
Add Mnist Pipeline for on-prem cluster

jinchihe:
The PR is to add Mnist Pipeline for on-prem cluster, includes:
1. `README` file to show how to run the Mnist Pipeline for on-prem cluster.
2. The Mnist Pipeline python classes definition file `mnist_pipeline_on_prem.py`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/519)


sarahmaddox:
@DanSanche Please would you review this PR, as it's related to your MNIST Pipelines for GCP. Also, I'd recommend that you submit a PR adding yourself to OWNERS for kubeflow/examples and kubeflow/examples/pipelines
 /cc @karthikv2k

DanSanche:
@sarahmaddox I'm a little unsure what to do here, because this is very similar to [the branch](https://github.com/DanSanche/examples/blob/mnist-pipelines/mnist-pipelines/mnist-pipeline.py) we were working on to form the basis of the web page. The differences being that this doesn't seem to rely on GCP resources, and this one doesn't seem to include the web UI. Do we really want to maintain two mnist-pipeline samples that are so similar?

I'm thinking either 1) I should stop work on mine, in which case you can work with @jinchihe to add in the web-ui and build the docs page based on this one, or 2) we close this PR and continue with my branch instead.

Thoughts?

pipelines/mnist/model.py
```diff
@@ -0,0 +1,207 @@
+#  Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
DanSanche: Is this file any different from the one in the mnist example? I think we should avoid having copies of the same code in multiple places for maintenance reasons

pipelines/mnist/mnist_pipeline_on_prem.py
```diff
@@ -0,0 +1,63 @@
...
+import kfp.dsl as dsl
+from kubernetes import client as k8s_client
+
+
+def mnist_train_op(tf_data_dir: str, tf_model_dir: str,
+                   tf_export_dir: str, train_steps: int, batch_size: int,
+                   learning_rate: float, step_name='Training'):
+    return dsl.ContainerOp(
+        name=step_name,
+        image='docker.io/jinchi/mnist_model:0.2',
```
DanSanche: Shouldn't this be TRAINING_IMAGE_NAME?

I'm not too sure if it's actually useful to use a parameter here though. I'd say we should just use an image at gcr.io/kubeflow-examples/mnist/model instead of a custom one, so people can jump into using the pipeline easier

pipelines/mnist/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+You can check the pipeline status from user interface on the experiment's details page. The following graphic shows the status of each component. 
+
+![Pipeline](mnist_pipeline.png "Mnist Pipeline")
+
+### Change the service type to NodePort
+
+After the Pipeline finished, The TF Serving will be created as a service. Change the service type to NodePort, so you can get a prediction from the service, as shown in the following:
+```
+kubectl -n kubeflow patch service mnist -p '{"spec": {"type": "NodePort"}}'
```
DanSanche: Is this something we could set in the pipeline instead of patching after?

sarahmaddox:
Thanks @DanSanche my recommendation is to close this PR and continue with yours. I've already done significant work in PR https://github.com/kubeflow/website/pull/490 based on your PR, and your PR is earlier than this one.

Currently, to the best of my knowledge, Kubeflow Pipelines doesn't support non-GCP installations, though the intention is to fix that. The issues that need fixing are:

* https://github.com/kubeflow/pipelines/issues/345
https://github.com/kubeflow/pipelines/issues/337

pipelines/mnist/model.py
jinchihe: Yes, you are right. I copied the model.py and updated that to let read args from command line instead of env variables. And now, that can be removed since the PR has been merged.

pipelines/mnist/mnist_pipeline_on_prem.py
jinchihe: You are absolutely right. I think that's better to public image 
gcr.io/kubeflow-examples/mnist/model, but I not sure if that's updated after some code merged.

pipelines/mnist/README.md
jinchihe: I has not idea for this, to get this, I think we need to update kubeflow pipeline deployer component, so that the service will be Nodeport while creating.

jinchihe:
@DanSanche @sarahmaddox Great thanks for your comments! Anyway, I updated the code change to address @DanSanche comments.

I respect your choice, but here personally I suggest to pay more attention for on-prem cluster cases. To let kubeflow more popular, we should support more case and add more examples for common k8s cluster which can use local storage.

I noticed the mnist example support 3 types: local storage, using GCS and using S3, is that better way to support more types for the mnist pipeline line example? Thanks.

/cc @jlewi

DanSanche:
@jinchihe Is it alright with you if I try to incorporate some of your code here onto my branch? I agree it would be ideal if we can make on-prem work for this example if possible, but it might be difficult to merge both branches individually

jinchihe:
> individually

Sounds good. Go ahead please if possible. Please let me know if need my help for somethings. Thanks. @DanSanche

PR #528, opened 2019-03-15
Enhance mnist training and add serving steps for local mode

jinchihe:
Tried the e2e mnist cases on the kubeflow on-prem cluster, but found two problems:
1. Under local mode, the training pod is not mount with local storage(pv/pvc), that’s useless because the directory is lost as soon as the pod is finished/deleted. We should save the model for serving.
2. No steps for serving the model for local mode. That’s TBD.

The PR is to:
1. Enhance the training component to associate the PVC to storage the trained model.
2. Add steps for serving the model for local storage mode.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/528)


jinchihe:
/assign @jlewi

Could please review the PR, to assign to someone? Thanks a lot!

mnist/README.md
```diff
@@ -722,9 +731,40 @@ kubectl describe service mnist-service
 
 TODO: Add instructions
 
-### PVC
+### Local storage
 
-TODO: Add instructions
+The section shows how to serve the local model that was storaged in PVC while training.
```
jlewi: storaged -> stored

jinchihe: Updated. Thanks @jlewi

PR #536, opened 2019-04-01
updated Dockerfile to work with new notebook_controller

zabbasi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/536)


zabbasi:
fixes https://github.com/kubeflow/kubeflow/issues/2680

zabbasi:
@abhi-g

zabbasi:
CMD/ENTRYPOINT is inherited from parent base image. No need to reset it.

PR #546, opened 2019-04-26
drop_ksonnet_from_mnist

### LINKED ISSUE #539: Remove ksonnet from mnist example
Since ksonnet is deprecated we should remove ksonnet from the mnist example.

One possible approach would be to use a tool like make, jsonnet, kustomize, or a custom python script to easily generate a set of YAML files customized for a particular user.

jinchihe:
fixes #539

## Problem: 

Remove `ksonnet` from mnist example since `ksonnet` is deprecated. 

### Overall: 

Using `kustomize` to replease `ksonnet`.

### For different platforms: Local, GCP and S3

There are three platforms we need to cover: Local, GCP and S3, for training, monitoring and serving. some configurations are common for the 3 platforms, while others are different (for example: need to configure pvc mounting for local mode and need to configure user-gcp-sa secret for GCP, etc..), the initial thought is creating `base`(common) and `overlay` for those differences, but seems the `kustomize` cannot not merge the CRD (`TFJOB`) correctly. For example: `base` configured common part, and `overlay` configured the specific volume configuration, but the kustomize `patchesStrategicMerge` works abnormally for the CRD, but that's OK for k8s native resource.

So the PR separated 3 platforms without `base`, that caused some duplication. seems kustomize need to be handle the CRD tfjob case, and then enhance the `mnist` to `base + overlay` methods.

### For customized configurations

There are some customized configurations (for example `modelDir` and `exportDir`), the method for handling the customized configurations is creating configmap for those customized configurations and then set variables from the `configmap`, kustomize will replace the variables with real value when general the YAML file by running `kustomize build`.

But here it is worth mentioning for two parameters: `numPs` and `numWorkers`, we cannot handle the parameters as above methods, two reasons:

1> From `configmap` -> variables solution, the value will be String (as "1"), but the CRD defined the value be integer（as 1).
2> If user do not want to set the Ps and Worker, just only chief. We cannot set the numPs and numWorkers to 0 since the CRD defined the values must be an integer and greater than or equal to 1. So the solution for `numPs` and `numWorkers` is separating `Chief`, `PS` and `Worker` to 3 YAML, if user defined `numPs` and `numWorkers` >= 1, then merge that, if not only defining `Chief` YAML in `resource` of `kustomization.yaml`.

I created one simple script `definition.sh` to get user customized configurations and then add resource in `kustomization.yaml` to merge Ps and Worker.

Any comments or concern, please let me know, thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/546)


jinchihe:
/assign @jlewi 

Jeremy, could you please review the RP, I will address your comments with high priority if have since you said that planed to run mnist demo by Apr 30. Thanks a lot!

jlewi:
Thanks @jinchihe 

/assign @swiftdiaries 
/assign @hougangliu 

Can one or both of you take a look as well since you are the kustomize experts?

jlewi:
@jinchihe Can you update the PR description please to provide more explanation in particular

1. What is the problem?
1. How you are addressing it?

jinchihe:
> @jinchihe Can you update the PR description please to provide more explanation in particular
> 
>     1. What is the problem?
> 
>     2. How you are addressing it?

Great thanks for your comments. Yes, that’s really need some explanation for the implement. I added  in the RP description, hope that’s help for reviewing the changes  :-)

mnist/training/local/kustomization.yaml
```diff
@@ -0,0 +1,36 @@
...
+    apiVersion: v1
+    kind: ConfigMap
+    name: mnist-map-local
+
+configurations:
+- params.yaml
+
+configMapGenerator:
+- literals:
+  name: mnist-map-local
```
hougangliu: syntax error here if use don't run `kustomize edit add configmap mnist-map-local --from-literal=name=mnist-train-local`. maybe you can set a default value here to avoid misunderstanding like below:
```
- name: mnist-map-local
  literals:	
  - name=mnist-train-local
```
please also update others similar as here

hougangliu:
as @jlewi mentioned, we had better have below org since mnist here is an app
.
├── kustomization.yaml
├── front
│   ├── kustomization.yaml
│   ├── ...
├── monitoring
│   ├── base
│   │   ├── kustomization.yaml
│   │   ├── ...
│   ├── GCS
│   │   ├── kustomization.yaml
│   │   ├── ...
│   └── S3
│   │   ├── kustomization.yaml
│   │   ├── ...
├── serving
│   ├── base
│   │   ├── kustomization.yaml
│   │   ├── ...
│   ├── GCS
│   │   ├── kustomization.yaml
│   │   ├── ...
│   └── local
│   │   ├── kustomization.yaml
│   │   ├── ...
└── training
    ├── base
    │   ├── kustomization.yaml
    │   ├── ...
    ├── GCS
    │   ├── kustomization.yaml
    │   ├── ...
    ├── local
    │   ├── kustomization.yaml
    │   ├── ...
    └── S3
    │   ├── kustomization.yaml
    │   ├── ...

mnist/training/local/kustomization.yaml
jinchihe: Thanks Hougang, if set default value, the current kustomize cannot set to new value for configmap, only support to add one metrics.

jinchihe:
> as @jlewi mentioned, we had better have below org since mnist here is an app
> .
> ├── kustomization.yaml
> ├── front
> │ ├── kustomization.yaml
> │ ├── ...
> ├── monitoring
> │ ├── base
> │ │ ├── kustomization.yaml
> │ │ ├── ...
> │ ├── GCS
> │ │ ├── kustomization.yaml
> │ │ ├── ...
> │ └── S3
> │ │ ├── kustomization.yaml
> │ │ ├── ...
> ├── serving
> │ ├── base
> │ │ ├── kustomization.yaml
> │ │ ├── ...
> │ ├── GCS
> │ │ ├── kustomization.yaml
> │ │ ├── ...
> │ └── local
> │ │ ├── kustomization.yaml
> │ │ ├── ...
> └── training
> ├── base
> │ ├── kustomization.yaml
> │ ├── ...
> ├── GCS
> │ ├── kustomization.yaml
> │ ├── ...
> ├── local
> │ ├── kustomization.yaml
> │ ├── ...
> └── S3
> │ ├── kustomization.yaml
> │ ├── ...

I did like that because seems kustomize does not support to handle the CRD `tfjob` case by `base` + `overlay` methods. I will have more investigation for this for how to implement to avoid duplication. Thanks.

mnist/Dockerfile.kustomize
```diff
@@ -1,20 +1,20 @@
...
-RUN chmod +x /usr/bin/ks
+RUN curl -O -L https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64
+RUN mv kustomize_${KUSTOMIZE_VERSION}_linux_amd64 /usr/bin/kustomize
+RUN chmod +x /usr/bin/kustomize
 
 RUN curl -L  https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl  -o /usr/bin/kubectl
 RUN chmod +x /usr/bin/kubectl
 
-#ksonnet doesn't work without a kubeconfig, the following is just to add a utility to generate a kubeconfig from a service account.
+#kustomize doesn't work without a kubeconfig, the following is just to add a utility to generate a kubeconfig from a service account.
```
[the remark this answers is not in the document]
jinchihe: Removed this to avoid misunderstanding.

jinchihe:
Hi Reviewers:

I updated the training for Local mode to support configure number of PS and number of Worker, and support configure parameters for training, that's hard code before in ksonnet.  Thanks

jlewi:
Can you file an issue please about TFJob not working with strategic patch merge?

jinchihe:
> Can you file an issue please about TFJob not working with strategic patch merge?

Logged new issue in the `kustomize` project:
https://github.com/kubernetes-sigs/kustomize/issues/1027

Waiting for answer, may be have solution to work around. I'm also going to do more investigation for how to resolve that to avoid code duplication, I will updated code then if have :-)

By the way, for your another comments for set default value in configmap for some variables. I logged anther ticket for requirement that supporting set configmap value using kustomize so that user can set that optionally and using default value.
https://github.com/kubernetes-sigs/kustomize/issues/1028

jlewi:
@jinchihe I think its fine to try to resolve code duplication in a follow on PR or issue.

Only thing I'd like to see before merging this is can we use a single copy of the definition.sh?

jinchihe:
> @jinchihe I think its fine to try to resolve code duplication in a follow on PR or issue.
> 
> Only thing I'd like to see before merging this is can we use a single copy of the definition.sh?

@jlewi Sorry for the late reply due to holidays. I addressed your comments for `definition.sh` just now.  And I'm also keeping an eye on how to resolve the code duplication. I will update that once get a work around for kustomize `base` + `overlay`. Thanks.

jinchihe:
Seems we can use `patchesJson6902` of kustomise to avoid code duplication, instead of patchesStrategicMerge. I'm going to do that. Thanks.

jinchihe:
Seems the `kubeflow-examples-presubmit` failed due to #549 , need to check the automation test environment. Thanks.

jlewi:
@jinchihe Apologies for the slow reply and for the frustrating testing experience. I replied on #549 
As a short term fix I think you could just change the cluster to:  kf-v0-5-n04

jinchihe:
@jlewi Thank you very much. The testing was successful seems the NFS volume problem has been fixed. :-)

And your comments have been addressed, summary the last change as following:
1. Used `patchesJson6902` of kustomise to avoid code duplication, now that's `base` + `overlay` model.
2. The script `definition.sh`, just have a single copy now.
3. As your comments, merged `command` and `args` in training containers, just only `command` now.  
4. Other micro enhancements.

Thanks a lot!

jlewi:
Woo Hoo! Thanks for your persistence!

PR #552, opened 2019-05-11
Add memory limit

texasmichelle:
This aids GKE autoprovisioning in creation of appropriate node sizes

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/552)


amygdala:
(don't think I'm an approver, but let me try)
/approve

PR #555, opened 2019-05-13
[docs]: enhance documentation for chosen namespace

travishen:
The web-ui pod is deploy on namespace `kubeflow`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/555)

### DIFF
--- a/mnist/README.md
+++ b/mnist/README.md
@@ -91,7 +91,9 @@ With our data and workloads ready, now the cluster must be prepared. We will be 
 
 In the following instructions we will install our required components to a single namespace.  For these instructions we will assume the chosen namespace is `kubeflow`.
 
-
+```
+kubectl config set-context $(kubectl config current-context) --namespace=kubeflow
+```
 
 ### Training your model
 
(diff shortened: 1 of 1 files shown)


travishen:
I signed it!

jinchihe:
@travishen In fact, we have description in the `Preparing your Kubernetes Cluster` section in the README file as `In the following instructions we will install our required components to a single namespace. For these instructions we will assume the chosen namespace is kubeflow`. so all commands in the README file omit the `-n kubeflow`. do you think is that need to clear this? if so, personally I suggest to add more clear descrition to remind user to configure k8s content to set `kubeflow` as current namespace. Thanks.

travishen:
@jinchihe Thanks for the remind. I've updated it by:
* Add command to set the namespace preference
* Omit the `-n kubeflow` in two commands

jinchihe:
> @jinchihe Thanks for the remind. I've updated it by:
> 
>     * Add command to set the namespace preference
> 
>     * Omit the `-n kubeflow` in two commands

Great! Thanks!

travishen:
Thanks for the review!

PR #559, opened 2019-05-17
Examples for pipelines running on-premise

### LINKED ISSUE #548: Examples for pipelines running on-premise

jinchihe:
fixes #548 

There is a requirement of on-prem preject, that's #548. I merged the pipeline for GCP and on-premise cluster. Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/559)


DanSanche:
I'm not too familiar with volumes in KFP yet, will your volume request work with GCP as well? If so, it would probably be worth using the same pipeline for both

jinchihe:
@DanSanche Thanks. I do not think the volume request work with GCP, but I think there is no impact for GCP sample, since the `--function` option (see updated Readme) will separate pipelines (workflow) for GCO and on-prem cluster, in other words, that are two ways for GCP and On-prem after compiled.

- For GCP
```
python venv/bin/dsl_compile --py mnist_pipeline.py --output mnist_pipeline_gcp.tar.gz --function mnist_pipeline_gcp
```
- For on-prem

```
python venv/bin/dsl_compile --py mnist_pipeline.py --output mnist_pipeline_onprem.tar.gz --function mnist_pipeline_onprem
```

jinchihe:
/hold

Will update more to try merge gcp and onprem.

jlewi:
Thanks @jinchihe 

@DanSanche I believe using PVC should work on GCP or any platform. I think the issue we run into with PVC is whether the PVC supports RW Many  or not. If it doesn't support RW many then you won't be able to run multiple steps in parallel that need to mount it.

One portable solution is to create a PVC for an NFS share. There are ways to run NFS on prem and cloud (including in cluster). 

That said for GCP its probably better to use GCS if possible then Cloud NFS (for reasons of cost and performance).

My bigger question is whether the PVC should be created as part of the pipeline?

The pattern I would usually follow is to create a PVC outside the pipeline and then just mount it if and when it was needed.

If you create the PVC inside the pipeline
  
  * What is the lifecycle of the PVC? Is the PVC deleted after the pipeline runs?
  * Is a new PVC created on each run?

The pattern I would follow (e.g [this notebook](https://github.com/jlewi/kubecon-demo/blob/master/ames-xgboost-build-train-deploy.ipynb)) is

   * Create a shared filesystem (e.g. Cloud Filestore on GCP)
   * Create a PVC for that filesystem
   * Mount that PVC on any job, pipeline, notebook that wants to share data

   * For example if I run a pipeline that uses that PVC and I also mount that PVC in my notebook then I can easily access any data in my notebook created by the pipeline

I think the main thing is I wouldn't want the lifecycle of the PVC to be tied to ephemeral objects like pipelines


/cc @IronPan 

p.s I'm at Kubecon this week and vacation next week so I will be very slow with reviews.

jinchihe:
Thanks your comments @jlewi and @DanSanche, Sorry for later :-)

Updated the code change as following, referred to implement method of [tfx sample](https://github.com/kubeflow/pipelines/tree/master/samples/tfx) in pipelines repo:
1. Addressed @jlewi comments, removed the PVC creating code, and let use input the pvc name.
2. For easy maintenance and to avoid code duplication, merged the pipeline definition functions for GCP and On-Prem cluster.
3. Define the variable platform to switch GCP and On-Prem, by default that's for GCP, if use want to try the pipeline on On-Prem cluster, need to update simplely as README description.
4. Updated Readme file, but keep the code part of readme file since I think that's introduction for user, that's clear.

jlewi:
Thanks!
/lgtm
/approve

bruceche11:
@jinchihe 
I run the example of  examples/pipelines/mnist/mnist_pipeline_on_prem.py  on my kubeflow,,,failed.
so I walk into the two Dockerfile,
https://github.com/kubeflow/examples/blob/master/mnist/Dockerfile.model
https://github.com/kubeflow/pipelines/blob/master/components/kubeflow/deployer/Dockerfile
am I walked wrong? if not, i continue
i trained succeed, but deploy, failed.
my question:
in the second Dockerfile(for deploy),there are google cloud sdk , ksonnet, k8s,
      ① the ksonnet is archived, so is it useful? 
      ②the k8s version must be 1.11.2 and the kubeflow version must be 0.4.0, right?

Thank you very much in advance.

jinchihe:
@bruceche11 I think the example is not updated recently, not sure if that can be executed with new kubeflow, but that should be fine with some enhancement. That's very great if you have time to enhance that. Thanks.

PR #560, opened 2019-06-01
Fixes to XGBoost example

richardsliu:
- Makes test_size, shuffle, and early_stopping_rounds configurable.
- Export model file only if specified.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/560)

### DIFF
--- a/xgboost_ames_housing/housing.py
+++ b/xgboost_ames_housing/housing.py
@@ -21,7 +21,7 @@ from sklearn.model_selection import train_test_split
 from sklearn.preprocessing import Imputer
 from xgboost import XGBRegressor
 
-def read_input(file_name, test_size=0.25):
+def read_input(file_name, test_size=0.25, shuffle=True):
   """Read input data and split it into train and test."""
   data = pd.read_csv(file_name[0])
   data.dropna(axis=0, subset=['SalePrice'], inplace=True)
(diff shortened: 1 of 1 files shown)


xgboost_ames_housing/housing.py
```diff
@@ -21,7 +21,7 @@
 from sklearn.preprocessing import Imputer
 from xgboost import XGBRegressor
 
-def read_input(file_name, test_size=0.25):
+def read_input(file_name, test_size=0.25, shuffle=True):
```
zhenghuiwang: Can we make shuffle=False as the default value?

It makes the model output deterministic and some integration test may rely on it.

xgboost_ames_housing/housing.py
```diff
@@ -103,14 +108,20 @@ def main(args):
...
           '--test-size',
           help='Fraction of training data to be reserved for test',
+          type=float,
           default=0.25
           )
+  parser.add_argument(
+          '--shuffle',
+          help='Whether to shuffle input before splitting',
+          type=bool,
+          default=True
```
zhenghuiwang: can we make the default value equals false?

jlewi:
@richardsliu Any update on this PR?

richardsliu:
I think this is no longer needed.

Issue #561, opened 2019-06-03
Invalid secret value "$(xxx)"


benjaminhuo:
When I test the mnist example, I got below errors when I apply the yaml generated by "kustomize build .":

```shell
tf-operator  Error creating: Pod "mnist-train-dist-worker-0" is invalid: 

[spec.containers[0].env[11].valueFrom.secretKeyRef.name: Invalid value: "$(awsSecretName)":

 a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), 

spec.containers[0].env[11].valueFrom.secretKeyRef.key: Invalid value: "$(awsAccessKeyIDName)": 

a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```

And the error disappears after I replace the $() with the actual value.

The yaml generated is attached

[kf-mnist.yaml.zip](https://github.com/kubeflow/examples/files/3246789/kf-mnist.yaml.zip)

jinchihe:
@benjaminhuo Thanks for trying the case for S3. Seems this is a bug, I'm going to fix the issue.

jinchihe:
@benjaminhuo Created a PR to fix the issue, you can apply the patch and retry again. Thanks.

benjaminhuo:
@jinchihe thanks for the quick response 👍, I'll apply the patch and try again

PR #566, opened 2019-06-05
fix issue for s3 case in mnist example

jinchihe:
fixes: #561

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/566)


jinchihe:
The `kubeflow-examples-presubmit` failed due to the issue #567 which has been fixed in PR #568 , thanks.

Issue #567, opened 2019-06-05
Upgrade tfjob version in the mnist example


jinchihe:
I installed last `kubeflow` and found the CRD tfjobs version has been upgraded.

```bash
#kubectl describe crd tfjobs.kubeflow.org
......
  Version:                      v1
  Versions:
    Name:     v1
    Served:   true
    Storage:  true
    Name:     v1beta2
    Served:   true
    Storage:  false
....
```

That's better to upgrade the tfjobs version to `v1` in the mnist example, otherwiase will print below error message while applying the exmaple 
```bash
[root@jinchi1 S3]# kustomize build . |kubectl apply -f -
configmap/mnist-map-training-5hb8dfd8kg created
secret/aws-creds-hk7k7f7f78 created
error: unable to recognize "STDIN": no matches for kind "TFJob" in version "kubeflow.org/v1beta1"
```

jinchihe:
I found the automation tests kubernetes logo
`kubeflow-examples-presubmit`  also faild because of this. 

```
error: unable to recognize "STDIN": no matches for kind "TFJob" in version "kubeflow.org/v1beta1"
ERROR|2019-06-05T06:51:54|/mnt/test-data-volume/kubeflow-examples-presubmit-mnist-566-dd56e69-2464-bf2b/src/kubeflow/tf-operator/py/kubef
low/tf_operator/test_runner.py|72| There was a problem running the job; Exception Command 'kustomize build /mnt/test-data-volume/kubeflow
-examples-presubmit-mnist-566-dd56e69-2464-bf2b/src/kubeflow/examples/mnist/training/GCS| kubectl apply -f -' returned non-zero exit stat
us 1
Traceback (most recent call last):
 File "/mnt/test-data-volume/kubeflow-examples-presubmit-mnist-566-dd56e69-2464-bf2b/src/kubeflow/tf-operator/py/kubeflow/tf_operator/tes
t_runner.py", line 53, in run_test
   test_func()
 File "tfjob_test.py", line 96, in test_train
   subprocess.check_call(subCmd, shell=True)
 File "/usr/lib/python2.7/subprocess.py", line 541, in check_call
   raise CalledProcessError(retcode, cmd)
CalledProcessError: Command 'kustomize build /mnt/test-data-volume/kubeflow-examples-presubmit-mnist-566-dd56e69-2464-bf2b/src/kubeflow/e
xamples/mnist/training/GCS| kubectl apply -f -' returned non-zero exit status 1
```

PR #568, opened 2019-06-05
update TFJob version to v1beta2

jinchihe:
fixes: #567 

Since the `TFJob` version has been upgraded in the last `kubeflow`, we need to update the `TFJob` version in the mnist example.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/568)

### DIFF
--- a/mnist/README.md
+++ b/mnist/README.md
@@ -295,7 +295,7 @@ To use this service account we perform the following steps
         kustomize build .
        ```
        ```
-        apiVersion: kubeflow.org/v1beta1
+        apiVersion: kubeflow.org/v1beta2
         kind: TFJob
         metadata:
           ...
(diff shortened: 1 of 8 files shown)


jinchihe:
/assign @jlewi 

Please help to review. Thanks a lot!

jlewi:
Are you sure you installed the v0.5.1 manifests? The CR version in v0.5.1 should not be V1.

It will be V1 in 0.6.

jinchihe:
@jlewi Sorry for the mistake, my test env was installed by `kfctl` and did not specify version information by`--version`, so that's master :-)

I think we must upgrade that since the automation tests `kubeflow-examples-presubmit` also failed because of no `v1beta1`. Maybe we can upgrade to `v1beta2` currently to work around this.

jinchihe:
Since both of last official release and `master` have `v1beta2` for TFJob, include version that is used by `kubeflow-examples-presubmit` , so update to `v1beta2` now.

Thanks  @jlewi

PR #569, opened 2019-06-05
Reusable Pipeline Steps for NLP

axsaucedo:
# Reusable Pipeline Steps for NLP
This PR contains an end-to-end example showcasing how to build a pipeline with re-usable containers that can pass artifacts using a volume. The last step in the pipeline deploys a Seldon graph using the same reusable components by attaching the volume containing the trained model biaries.

The example can be followed by running the commands provided in the [Jupyter notebook](https://github.com/axsauze/examples/blob/reusable_components_example/reusable_components_nlp/kubeflow_seldon_e2e_pipeline.ipynb), or just following the instructions on the  [README](https://github.com/axsauze/examples/tree/reusable_components_example/reusable_components_nlp)

The Kubeflow pipeline built in this example looks as follows:

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/569)


axsauze:
The CLA has been signed already on an employer level by [Seldon](https://seldon.io), which also covers this PR. Please let me know if there are any changes required to make sure it aligns with the requirements of the Kubeflow examples.

texasmichelle:
I submitted a request for your company's POC so that you can double check the email in your GH account is listed as an authorized contributor.

There are a large number of files in this PR, many of which are in .cache directories. I created #573 to ignore that dir, which needs to be merged first.

texasmichelle:
It looks like the commit history is associated with a gmail address. @ahousley is the POC and either he needs to add your gmail to the corporate group or you need to update the commit history to reflect your seldon.io email (and make sure it's added to your GitHub account).

axsauze:
Thank you very much for the heads up @texasmichelle. Your comment made me realise that the folder you pointed out should not be in there on the first place, as it's created on the [first step in the tutorial](https://github.com/axsauze/examples/tree/reusable_components_example/reusable_components_nlp), so the PR is no longer dependent on #573. I also realised that my `.mypy_cache` folder was also part of the commit, but I've changed that and added the folder to the `.gitignore` file and have squashed all commits. As per your request, I have also changed the repository permissions so that all relevant commits are on the as@seldon.io address - it seems that CLA/Google requirements are approved now.

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+    seldon_core
+    numpy
+
+
+
+```python
+!pip install -r requirements-dev.txt
+```
+
+## 1) Run all the services (Kubeflow and Seldon)
```
texasmichelle: This could use some clarity that it will install default Kubeflow components on an existing cluster. Include an escape hatch for users that already have KF installed.

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+kfctl apply all -V
+```
+
+### Now let's run Seldon 
+For this we'll need Helm to be running, so we'll initialise it.
+
+
+```bash
+%%bash
+helm init
```
texasmichelle: Rather than introduce another packaging tool, let's use kustomize here (ksonnet if necessary, but since it's being retired I'd rather not use it for net-new code).

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+
+
+```python
+!kubectl get pod -n kubeflow | grep seldon
+```
+
+    seldon-operator-controller-manager-0                       1/1     Running   1          6s
+
+
+### Temporary fix for Argo image
```
texasmichelle: Please ensure there are outstanding issues for these fixes in Core & Pipelines repos (or wherever the permanent fix belongs)

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+!kubectl -n kubeflow patch deployments. workflow-controller --patch '{"spec": {"template": {"spec": {"containers": [{"name": "workflow-controller", "image": "argoproj/workflow-controller:v2.3.0-rc3"}]}}}}'
+!kubectl -n kubeflow patch deployments. ml-pipeline --patch '{"spec": {"template": {"spec": {"containers": [{"name": "ml-pipeline-api-server", "image": "elikatsis/ml-pipeline-api-server:0.1.18-pick-1289"}]}}}}'
+# !kubectl -n kubeflow patch configmaps workflow-controller-configmap --patch '{"data": {"config": "{ executorImage: argoproj/argoexec:v2.3.0-rc3,artifactRepository:{s3: {bucket: mlpipeline,keyPrefix: artifacts,endpoint: minio-service.kubeflow:9000,insecure: true,accessKeySecret: {name: mlpipeline-minio-artifact,key: accesskey},secretKeySecret: {name: mlpipeline-minio-artifact,key: secretkey}}}}" }}'
+```
+
+    deployment.extensions/workflow-controller patched
+    deployment.extensions/ml-pipeline patched
+
+
+The last command you need to run actually needs to be manual as the patch cannot change configmap contents directly
```
texasmichelle: Consider providing a file with this change that the user can apply

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+# we must be in the same directory
+cd pipeline/pipeline_steps/clean_text/ && ./build_image.sh
+cd ../data_downloader && ./build_image.sh
+cd ../lr_text_classifier && ./build_image.sh
+cd ../spacy_tokenize && ./build_image.sh
+cd ../tfidf_vectorizer && ./build_image.sh
+```
+
+## 3) Train our NLP Pipeline through the Kubeflow UI
+We can access the Kubeflow dashboard to train our ML pipeline via http://localhost/_/pipeline-dashboard
```
texasmichelle: How about instructing users to connect to their dashboard in the usual way with an external URL or giving them a port-forward command?
`kubectl -n kubeflow port-forward svc/ambassador 8080:80`

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+    }
+    
+
+
+## 6) Visualise Seldon's Production ML Pipelines
+We can visualise the performance using the SeldonAnalytics package, which we can deploy using:
+
+
+```python
+!helm install seldon-core-analytics --repo https://storage.googleapis.com/seldon-charts --namespace kubeflow
```
texasmichelle: Same comment about using kustomize

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+```
+
+In my case, similar to what I did with Ambassador, I need to make sure the the service is a LoadBalancer instead of a NodePort
```
texasmichelle: Same comment about port-forwarding

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+while True:
+    client_prediction = sc.predict(
+        data=batch, 
+        deployment_name=deployment_name,
+        names=["text"],
+        payload_type=payload_type,
+        transport="rest")
+```
+
+## You now have a full end-to-end training and production NLP pipeline 😎
```
texasmichelle: Perhaps briefly explain what the graphs show here

texasmichelle:
Adding @vicaire for his pipelines expertise (feel free to swap out)

axsauze:
@texasmichelle thank you for the comments, I'll go through each point in detail and resolve it. In regards to the suggestion you are completely right, we actually demoed this in a conference successfully where we used various datasets being passed as inputs, including a twitter sentiment analysis dataset, and we created a load balance tester container that continuously sends tweets to the deployed model on a real time search query to showcase the performance and real time metrics as the new model is replaced. Great stuff, looking forward to hear @vicaire's thoughts as well.

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+
+5) Test Seldon Deployed ML REST Endpoints
+
+6) Visualise Seldon's Production ML Pipelines
+
+# Before you start
+Make sure you install the following dependencies, as they are critical for this example to work:
+
+* Helm v2.13.1+
+* A Kubernetes cluster running v1.13 or above (minkube / docker-for-windows work well if enough RAM)
```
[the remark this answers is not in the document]
axsauze: You're correct, it was renamed

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
+
+# End-to-end Reusable ML Pipeline with Seldon and Kubeflow
+
+In this example we showcase how to build re-usable components to build an ML pipeline that can be trained and deployed at scale.
+
+We will automate content moderation on the Reddit comments in /r/science building a machine learning NLP model with the following components:
+
+![](img/completed-pipeline-deploy.jpg)
```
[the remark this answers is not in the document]
axsauze: I don't think I understand what you mean by being difficult to see the components. Is it because the photo is long and doesn't fit in the screen? The sections below provide cropped versions of the image to show the relevant pieces whilst this picture is intended to provide a full overview.

reusable_components_nlp/README.md
```diff
@@ -0,0 +1,989 @@
...
+    https://storage.googleapis.com/ml-pipeline/release/0.1.20/kfp.tar.gz
+    kubernetes
+    click
+    seldon_core
+    numpy
+
+
+
+```python
+!pip install -r requirements-dev.txt
```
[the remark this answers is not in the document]
axsauze: No, this is a `jupyter nbconvert` to README.md, the notebook is provided in the folder, I could add a note at the top pointing this out, alternatively I can only leave the intro and point a link to the notebook instead of leaving the full contents in `.md` format.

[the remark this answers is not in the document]
axsauze: Or do you mean specifically in regards to this command? If that is the case yeah I can just change that to a markup type and suggest for people to run it in the terminal.

reusable_components_nlp/README.md
axsauze: I don't understand what you mean. What is an escape hatch? Perhaps if there are other examples that have this outlined I could try to add the same terminology when they install kubeflow. It should be quite self explanatory tho, if they already have KF installed they should skip the installation of Kubeflow.

reusable_components_nlp/README.md
axsauze: I see where you're coming from. At seldon we use helm as our default package manager tho, and the latest seldon charts are continuously updated in the helm repo. There is currently a deployment of Kubeflow that has a version of Seldon but the version is quite old and would need updating for that. What would you suggest here?

reusable_components_nlp/README.md
axsauze: Oh like providing the whole configmaps file? I had difficulties actually trying to do the change in a different way than an edit (hence why I'm not even using patch as previously). This unfortunately is a quickfix that should will hopefully be avoided after kubectl upgrades to the next version of argo.

reusable_components_nlp/README.md
axsauze: Fair enough

reusable_components_nlp/README.md
axsauze: Same issue as mentioned above

reusable_components_nlp/README.md
axsauze: 👍

reusable_components_nlp/README.md
axsauze: 👍

reusable_components_nlp/README.md
texasmichelle: This is a common pain point & we haven't worked out a great way to accommodate it yet. The motivation for this request is to reduce the user requirement for additional tooling. The other point you brought up is that KF currently ships an out-of-date version of Seldon, which isn't what we want either. I think it's worthwhile to update the component that ships with KF releases. I'm unopinionated on which occurs first, but the same commands should be applicable both here and there.

texasmichelle: Also, I know this has been a blocker for integration of other components as well, so it's worth bringing up in slack or on a weekly call so that we can figure out the best approach.

reusable_components_nlp/README.md
texasmichelle: By escape hatch I just mean a brief statement that this section is optional if kubeflow is already installed on the cluster (the section refers to running the services, which is a bit ambiguous for new users). For users unfamiliar with `kfctl`, it is unclear which steps generate files and which ones modify the cluster. Being a bit more explicit about what is going on makes it smoother for new users.

reusable_components_nlp/README.md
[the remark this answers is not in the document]
texasmichelle: ah, I see. I'd rather have the instructions in only one place, so a link to the notebook instead of generated markdown would be better. GitHub renders notebooks cleanly so viewing isn't an issue anymore, but the downside is that version history is more obfuscated. I kind of like this approach with commands integrated, but it is different from all of our other examples. Is the expectation that a user is running a notebook server locally? Perhaps just mention that in the markdown.

reusable_components_nlp/README.md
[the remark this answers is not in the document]
texasmichelle: The wording of the previous sentence makes it seem like I should be paying attention to the list of components, which I can't read in the image. Referring to it as a preview instead would be fine.

texasmichelle:
A rebase should resolve the file conflicts

Issue #464, opened 2019-01-08
[mnist_pytorch] Remove specific modules from .pylintrc ignore-patterns


dsdinter:
Remove py,mnist_DDP.py and mnistddpserving.py modules from .pylintrc ignore-patterns , adding specific exceptions using inline comments instead.

We should still keep the autogenerated Python GRPC modules:
object_detection_grpc_client.py
prediction_pb2.py
prediction_pb2_grpc.py

texasmichelle:
Is this issue still outstanding?

dsdinter:
It is still outstanding. Will try to implement next week.
Thanks for the heads up!

PR #581, opened 2019-06-28
Lint fixes mnist

dsdinter:
Fixes: #464

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/581)

### DIFF
--- a/pytorch_mnist/ks_app/app.yaml
+++ b/pytorch_mnist/ks_app/app.yaml
@@ -1,8 +1,11 @@
 apiVersion: 0.3.0
+environments:
+  default:
+    destination:
+      namespace: default
+      server: https://104.154.168.244
+    k8sVersion: v1.8.0
+    path: default
 kind: ksonnet.io/app
-name: ks-app
-registries:
-  incubator:
-    protocol: github
-    uri: github.com/ksonnet/parts/tree/master/incubator
+name: ks_app
 version: 0.0.1
--- a/pytorch_mnist/training/ddp/mnist/mnist_DDP.py
+++ b/pytorch_mnist/training/ddp/mnist/mnist_DDP.py
@@ -17,15 +17,14 @@ limitations under the License.
 import datetime
 import logging
 import os
-import sys
 from math import ceil
 from random import Random
 
 import torch
 import torch.distributed as dist
-import torch.nn as nn
+import torch.nn as nn  # pylint: disable = useless-import-alias
 import torch.nn.functional as F
-import torch.optim as optim
+import torch.optim as optim # pylint: disable = useless-import-alias
 import torch.utils.data
 import torch.utils.data.distributed
 from torch._utils import _flatten_dense_tensors, _unflatten_dense_tensors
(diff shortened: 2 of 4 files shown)


dsdinter:
@texasmichelle please review and approve if OK for you.

texasmichelle:
/lgtm
Thanks @dsdinter! Curious why so much disabling?

dsdinter:
Thanks, good question. The Pytorch official example from where this one is based on has the same lint errors/warning, https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/distributed_cpu.py
Includes protected methods and using the forward method with different arguments.

There are some which could be fixed actually, e.g., more simpler imports

jinchihe:
@dsdinter Thanks. I have two quick questions.

1. Not sure where the python file is from, if from the 3rd-party, I think that's accept to ignore that.
2. I saw your removed the two files from ignore list, but disable all pylint errors in the file, without any enhancement/fix. Confused me here, seems that have sample efficacy for ignore and disabled all in files.

pytorch_mnist/training/ddp/mnist/mnist_DDP.py
```diff
@@ -234,10 +235,10 @@ def run(modelpath, gpu):
   args = parser.parse_args()
   if args.gpu:
     logging.info("\n======= CUDA INFO =======")
-    logging.info("CUDA Availibility:", torch.cuda.is_available())
-    if (torch.cuda.is_available()):
-      logging.info("CUDA Device Name:", torch.cuda.get_device_name(0))
-      logging.info("CUDA Version:", torch.version.cuda)
+    logging.info("CUDA Availibility:", torch.cuda.is_available())  # pylint: disable = logging-too-many-args
```
jinchihe: I think here will be have traceback confirm? also same below line.

texasmichelle:
@dsdinter can you address @jinchihe's questions?

pytorch_mnist/training/ddp/mnist/mnist_DDP.py
dsdinter: You are right, thanks for highlighting the typo.

dsdinter:
@texasmichelle @jinchihe let me know if there is anything else before we can merge this PR.

pytorch_mnist/ks_app/app.yaml
```diff
@@ -1,8 +1,11 @@
 apiVersion: 0.3.0
+environments:
+  default:
+    destination:
+      namespace: default
+      server: https://104.154.168.244
+    k8sVersion: v1.8.0
+    path: default
```
jinchihe: Thanks for the contribution!  Do we have to modify this file? since that works fine before after updated the `apiVersion` to "0.3.0".

jinchihe: Checked, seems that's same with `./examples/github_issue_summarization/ks_app/app.yaml`. I think the change is OK.

Issue #583, opened 2019-07-01
[codelab-intro-mnist-feedback]:


mitsuan:
kustomize build isn't working for 2.1.0

Getting error:
Error: var '{batchSize ~G_~V_ConfigMap {data.batchSize}}' cannot be mapped to a field in the set of known resources

jinchihe:
@mitsuan See details in https://github.com/kubeflow/examples/issues/578, the new version kustomize may changed something, I am going to enhance that to fit new kustomize version. but you can use v2.0.3 to work around.

Issue #578, opened 2019-06-25
MNIST Example Doesn't Work


Tabrizian:
I tried the local version of MNIST example and kustomize fails with the error below:

```
Error: var '{batchSize ~G_~V_ConfigMap {data.batchSize}}' cannot be mapped to a field in the set of known resources
```

jinchihe:
Could you please provide more information, such as kustomize version? k8s version, and which step the problem happened? @Tabrizian

jinchihe:
You can try to check the batchSize is set by running, or list the training/local/kustomization.yaml file here.

```
# grep -i batchSize kustomization.yaml
  - batchSize=100
```

davidtwomey:
@Tabrizian @jinchihe 
Was having the same issue using `kustomization` latest release (v3.0.0 pre) as per tutorial:
https://codelabs.developers.google.com/codelabs/kubeflow-introduction/#0

Got `kustomization build .` to work by
- Switching to explicit `kustomisation` v2.0.0
- Changing from `kustomize edit set image training-image=$TRAIN_PATH` to<br> `kustomize edit set image training-image=$TRAIN_PATH:latest`
i.e. appending image tag

> I tried the local version of MNIST example and kustomize fails with the error below:
> 
> ```
> Error: var '{batchSize ~G_~V_ConfigMap {data.batchSize}}' cannot be mapped to a field in the set of known resources
> ```

jinchihe:
@davidtwomey Yes, I tested kustomization 2.0.3, that should work fine.  And for you mentioned "$TRAIN_PATH:latest" ,  In the readme file, the $TRAIN_PATH should include tag as below description:
```
DOCKER_URL=docker.io/reponame/mytfmodel:tag
```

davidtwomey:
@jinchihe 
Yep - the $TRAIN_PATH was just a heads up for those following via the Google colabs tutorial
where train path did not have a tag
GCP `TRAIN_PATH=us.gcr.io/$PROJECT_ID/kubeflow-train`

Tabrizian:
```
kubectl version
[...]

PR #584, opened 2019-07-02
recomment using kustomize v2.0.3 for mnist

jinchihe:
Recently someones hit one problem while running mnist example with new kustomize (v2.1.0), such as #583 and #578,  after checked the problem, I think that's should be a issue in new kustomize, logged [ticket](https://github.com/kubernetes-sigs/kustomize/issues/1295) to trace in kustomize.

But avoid user stepping on the pit, the PR is to add note to recomment user to use kustomize 2.0.3. May enhance later if the kustomzie issue fixed.

fixes: #583 #578

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/584)

### DIFF
--- a/mnist/README.md
+++ b/mnist/README.md
@@ -50,6 +50,8 @@ You also need the following command line tools:
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
 - [kustomize](https://kustomize.io/)
 
++**Note:** kustomize [v2.0.3](https://github.com/kubernetes-sigs/kustomize/releases/tag/v2.0.3) is recommented since the [problem](https://github.com/kubernetes-sigs/kustomize/issues/1295) in kustomize v2.1.0.
+
 To run the client at the end of the example, you must have [requirements.txt](requirements.txt) installed in your active python environment.
 
 ```
--- a/test/workflows/components/mnist.jsonnet
+++ b/test/workflows/components/mnist.jsonnet
@@ -20,7 +20,7 @@ local defaultParams = {
   // Which Kubeflow cluster to use for running TFJobs on.
   kfProject: "kubeflow-ci-deployment",
   kfZone: "us-east1-b",
-  kfCluster: "kf-v0-5-n04",
+  kfCluster: "kf-vmaster-n01",
 
   // The bucket where the model should be written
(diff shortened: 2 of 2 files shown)


jinchihe:
Tested failed due to 
```
error: unable to recognize "STDIN": no matches for kind "TFJob" in version "kubeflow.org/v1beta2
```
Seems the TFJob is upgraded in the test cluster? will confirm later.

jinchihe:
That's stange, checked the code, the version v1beta2 of TFJob should exist in the cluster, check some code to show the TFJob version for debug... Thanks.

jinchihe:
From the debug information, the tfjobs is not installed in the testing cluster. As below error:
```
Error from server (NotFound): customresourcedefinitions.apiextensions.k8s.io “tfjobs.kubeflow.org” not found
```
@jlewi Do you know what's changed in the testing cluster? Anyone can help to check/fix this? Thanks.

jlewi:
What cluster are you using?
Is it kf-vmaster-XX
Those clusters are now deploying from master which might be using TFJob v1.

jlewi:
I checked v-kfmaster-n01
and it looks like TFJob is defined.

See attached
[tfjob.yaml.txt](https://github.com/kubeflow/examples/files/3360408/tfjob.yaml.txt)

Its v1.

jlewi:
Opened #588 to give you access to the clusters.

jinchihe:
@jlewi Many thanks. Yes, that works fine with the cluster!

aginpatrick:
@jinchihe @jlewi  would it be useful to add that kustomize v2 blocks relative paths (it's no more true with v3). So kustomization.yaml cannot contain "resources: -../bases" for example. I *think* that substitute resources for bases is the solution in this case.

aginpatrick:
@jinchihe @jlewi btw, does using kustomize v2.0.3 has implications for kubectl version? I mean, is v2.0.3 compatible with latest kubectl or one has to use older version?

jinchihe:
@aginpatrick kustomize should have no implications with kubectl version. But kustomize has been intergrated in the kubectl, see https://kubernetes.io/blog/2019/03/25/kubernetes-1-14-release-announcement/. kustomize v2.0.3 is available in kubectl v1.15.

PR #594, opened 2019-07-12
fix e2e test failed problem

jinchihe:
E2E tests always failed recent days, the the problem has been fixed in aother project under kubeflow, apply the fix solution to examples.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/594)


texasmichelle:
Thanks for this @jinchihe !!

PR #595, opened 2019-07-12
Update app.yaml to 0.3.0

richardsliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/595)


jinchihe:
@richardsliu Thanks, that's already updated in the https://github.com/kubeflow/examples/pull/594

/close

PR #603, opened 2019-07-18
Install pip dependencies and build base image with kaniko

### LINKED ISSUE #600: xgboost_synthetic exampe - run in stock TF notebook image; install pip dependencies and build base image with kaniko
This example:
https://github.com/kubeflow/examples/tree/master/xgboost_synthetic

Currently assumes you are running in a notebook image custom built for that example. We'd like to instead update the example so we can run it in a stock notebook image that ships with Kubeflow.

This needs we need to make the following changes

1. In our notebook we need to install all the pip packages that are required but not included in the base package. Here's a list

    ```
   !pip install sklearn
   !pip install pandas
   !pip install joblib
   !pip install fire
    ```

    * We should the code here https://github.com/kubeflow/examples/blob/ac9f2f12388c0e9d6260498b655e24faaabc71db/xgboost_synthetic/util.py#L14

1. We should use the [Kaniko builder](https://github.com/kubeflow/fairing/tree/master/fairing/builders/cluster) to build the base image from the Dockerfile in the repo

    * We should then use this as the base image with the append builder

abcdefgs0324:
- Install pip packages not included in the base package
- Use Kaniko builder to build the base image from the Dockerfile first

Fix #600 

/assign @jlewi 
/assign @zhenghuiwang

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/603)

### DIFF
--- a/xgboost_synthetic/util.py
+++ b/xgboost_synthetic/util.py
@@ -11,6 +11,8 @@ KFP_PACKAGE = 'https://storage.googleapis.com/ml-pipeline/release/0.1.20/kfp.tar
 def notebook_setup():
   # Install the SDK
 
+  for p in ["sklearn", "pandas", "joblib", "fire"]:
+    subprocess.check_call(["pip3", "install", p])
   subprocess.check_call(["pip3", "install", KFP_PACKAGE, "--upgrade"])
 
   logging.basicConfig(format='%(message)s')
--- a/xgboost_synthetic/util.py
+++ b/xgboost_synthetic/util.py
@@ -11,8 +11,7 @@ KFP_PACKAGE = 'https://storage.googleapis.com/ml-pipeline/release/0.1.20/kfp.tar
 def notebook_setup():
   # Install the SDK
 
-  for p in ["sklearn", "pandas", "joblib", "fire"]:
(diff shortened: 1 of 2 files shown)


xgboost_synthetic/util.py
```diff
@@ -11,6 +11,8 @@
 def notebook_setup():
   # Install the SDK
 
+  for p in ["sklearn", "pandas", "joblib", "fire"]:
```
abhi-g: xgboost_synthetic has a requirements.txt file that lists all the dependencies. Could we do pip3 install -r requirements.txt.

that way going forward only the requirements.txt needs updates.

abcdefgs0324: These packages have been included in requirements.txt.
This change is to help us be able to run it in a stock notebook image that ships with Kubeflow.

abhi-g: Correct. I meant that instead of iterating over these in the notebook, if we can pip3 install -r requirements.txt in here, that would future proof this. i.e. if new dependencies are added to requirements.txt we won't need to modify the code here.

abcdefgs0324: Oh, I got it. I will fix it.

PR #618, opened 2019-08-12
object_detection: fix typo error in tf-serving.libsonnet

MrXinWang:
modified tf-serving.libsonnet in object_detection example to fix the error of
"FileSystemStoragePathSource encountered a file-system access error:
Could not find base path /models/model for servable model"

Change-Id: I946a0a7fbb6c80992d66fe003ca90b1c21c67cfc
Signed-off-by: Henry Wang <henry.wang@arm.com>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/618)

### DIFF
--- a/object_detection/ks-app/vendor/kubeflow/tf-serving/tf-serving.libsonnet
+++ b/object_detection/ks-app/vendor/kubeflow/tf-serving/tf-serving.libsonnet
@@ -119,8 +119,10 @@
       name: $.params.name,
       image: $.params.modelServerImage,
       imagePullPolicy: "IfNotPresent",
-      args: [
+      command: [
         "/usr/bin/tensorflow_model_server",
+      ],
+      args: [
         "--port=9000",
         "--model_name=" + $.params.modelName,
         "--model_base_path=" + $.params.modelPath,


jinchihe:
@MrXinWang Could you please explain more why the code changes can fix the issue? Thanks.

/ok-to-test

MrXinWang:
@jinchihe of course.

According to the setup in [params.libsonnet](https://github.com/kubeflow/examples/blob/master/object_detection/ks-app/components/params.libsonnet) and [tutorial step of serving the model](https://github.com/MrXinWang/examples/blob/master/object_detection/tf_serving_cpu.md), the trained model is mounted to /mnt/exported_graphs/saved_model directory of the container "pets-model" created by image "tensorflow/serving:1.10.0". By checking the [dockerfile](https://github.com/tensorflow/serving/blob/1.10.0/tensorflow_serving/tools/docker/Dockerfile) of this image, the default model base path is set as "/models" and the default model name is set as "model". Therefore, using the previous tf-serving.libsonnet, tf_model_server will still try to find a servable model from default path and then throw the error.

From my understanding, coding everything in "args" label will probably cause wrong argument passing behaviour described above, because without the "command" label, the container will execute the default entrypoint. From the changed code, the container will execute "/usr/bin/tensorflow_model_server" command, which is a binary for serving the model, with parameters "--port=9000", "--model_name=" + $.params.modelName, and "--model_base_path=" + $.params.modelPath. This should be the expected behaviour.

I have tested this change on both my x86 and arm64 machine for this change and they both worked well. Please correct me if there is any mistake. Thanks.

jinchihe:
@MrXinWang Great, since I just saw that's from vendor..

Pls asign CLA as above reminder, thanks.

MrXinWang:
@jinchihe We are on it. Thank you very much.

jinchihe:
@MrXinWang Need a Contributor License Agreement (CLA) for the commit authors, pls sign CLA following above reminder. Thanks.

MrXinWang:
@jinchihe Hi, thanks for the reminding, as far as I know, I have been already added to the group of google project contributers from my company, which, I think, means I have already signed the CLA (company version). Should I re-propose a PR or is there a method to re-run the check of this PR? Thanks!

jinchihe:
@MrXinWang 
You need to visit https://cla.developers.google.com/ to sign.
Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!`)

MrXinWang:
@googlebot I signed it!

@jinchihe I think the check looks good now.

jinchihe:
@MrXinWang Yes, that's OK once googlebot get the message,. Thanks for the contribution.

/lgtm
/approve

Issue #619, opened 2019-08-16
Friction log - xgboost_synthetic on 0.6.2 RC


jlewi:
I tried running the xgboost example following the release demo script.
http://bit.ly/2ZgOE1v

The first problem I ran into is that the stock Jupyter image has an older version of fairing installed which doesn't have fire.

We should add a line to install from github

```
!pip3 install git+git://github.com/kubeflow/fairing.git@c6c075dece72135f5883abfe2a296894d74a2367
```

jlewi:
The next problem I hit is that instantiating the ModelServe class fails with a connection timeout. It looks like its trying to create a client for the metadata server.

```
---------------------------------------------------------------------------
RemoteDisconnected                        Traceback (most recent call last)
/opt/conda/lib/python3.6/site-packages/urllib3/connectionpool.py in urlopen(self, method, url, body, headers, retries, redirect, assert_same_host, timeout, pool_timeout, release_conn, chunked, body_pos, **response_kw)
    599                                                   body=body, headers=headers,
--> 600                                                   chunked=chunked)
    601 

/opt/conda/lib/python3.6/site-packages/urllib3/connectionpool.py in _make_request(self, conn, method, url, timeout, chunked, **httplib_request_kw)
    383                     # otherwise it looks like a programming error was the cause.
--> 384                     six.raise_from(e, None)
    385         except (SocketTimeout, BaseSSLError, SocketError) as e:

/opt/conda/lib/python3.6/site-packages/urllib3/packages/six.py in raise_from(value, from_value)

/opt/conda/lib/python3.6/site-packages/urllib3/connectionpool.py in _make_request(self, conn, method, url, timeout, chunked, **httplib_request_kw)
    379                 try:
--> 380                     httplib_response = conn.getresponse()
    381                 except Exception as e:

/opt/conda/lib/python3.6/http/client.py in getresponse(self)
   1330             try:
-> 1331                 response.begin()
   1332             except ConnectionError:

/opt/conda/lib/python3.6/http/client.py in begin(self)
    296         while True:
--> 297             version, status, reason = self._read_status()
    298             if status != CONTINUE:

/opt/conda/lib/python3.6/http/client.py in _read_status(self)
    265             # sending a valid response.
--> 266             raise RemoteDisconnected("Remote end closed connection without"
    267                                      " response")

RemoteDisconnected: Remote end closed connection without response

During handling of the above exception, another exception occurred:

ProtocolError                             Traceback (most recent call last)
[...]

PR #621, opened 2019-08-18
Fix a bunch of issues with the xgboost_synthetic example

jlewi:
* Need to add kfmd to requirements.txt because the training code now uses
  kfmd to log data.

* The Dockerfile didn't build with kaniko; it looks like a permission problem
  trying to install python files into the conda directory. The problem appears
  to be fixed by not switching to user root.

* Updte the base docker image to 1.13.

* Remove some references in the notebook to namespace because the fairing
  code should now detect namespace automatically and the notebook will no longer
  be running namespace kubeflow

* When running training in a K8s job; the code will now try to contact the
  metadata server but this can fail if the ISTIO side car hasn't started yet.
  So we need to wait for ISTIO to start; we do this by trying to contact
  the metadata server for up to 3 minutes.

* Add a lot more explanation in the notebook to explain what is happening.

* Related to #619

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/621)


zhenghuiwang:
/lgtm

Thanks for the fix!

PR #626, opened 2019-08-22
More KF06-related updates to the 'cookbook' branch notebooks

amygdala:
This is an update to the 'cookbook' branch, which contains some examples in progress.
These changes are mostly additional updates to get these notebooks in sync with the KF06 changes.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/626)


cookbook/pipelines/notebooks/README.md
```diff
@@ -4,6 +4,6 @@
 This directory contains several notebook-based examples of how to deploy a [Kubeflow Pipeline](https://www.kubeflow.org/docs/pipelines/) from outside the Kubeflow cluster.
 
 - [kfp_remote_deploy.ipynb](./kfp_remote_deploy.ipynb) shows how to do this with an IAP-enabled Kubeflow installation (on GKE). 
-- [kfp_remote_deploy-no-IAP.ipynb](./kfp_remote_deploy-no-IAP.ipynb) walks through how to connect via port-forwarding to the cluster
+- [kfp_remote_deploy-no-IAP.ipynb](./kfp_remote_deploy-no-IAP.ipynb) walks through how to deploy via a port-forwarding connection to a cluster.
```
jlewi: Does this example need to be updated now for 0.6? What method of deploying Kubeflow are you using for portforwarding?

amygdala:
I've updated that 'port-forwarding' example to an extent, but I think it's still a bit problematic/unclear. It references Kubeflow installations "set up, e.g. as described in https://www.kubeflow.org/docs/started/k8s/kfctl-k8s-istio/". 
That's probably not the entirely correct spin, though.  And there's also the lightweight pipelines install (https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize), which should get a mention.

In fact, for the purposes of this example— which talks only about remote pipeline deployment— the user could also **port-forward to the pipelines UI only**, which would work more broadly, I think for pretty much any type of installation:

```
kubectl port-forward -n kubeflow svc/ml-pipeline-ui 8080:80
```
Perhaps I should use that command in this example notebook -- WDYT?

(Unlike the other two notebooks, this one is not meant to be GCP-specific, and I was thinking it was useful to include as a general-purpose tactic... though the story is getting more complicated than it used to be).

amygdala:
/hold
on pulling this in just yet: I've decided to add some additional content, plus the issue above needs more clarity.

amygdala:
Jeremy, what do you think of this approach? https://github.com/kubeflow/examples/blob/415317dbefb84b95e544b9bd524375f16e250f16/cookbook/pipelines/notebooks/kfp_remote_deploy-port-forward.ipynb

The example now just uses port-forwarding to the Pipelines UI (which I believe should work across the board, making things simple), but notes that this is not providing access to the rest of the KF Dashboard.  It also gives a callout that in some configurations it might be possible to port-forward to the KF Dashboard itself.
@jlewi

jlewi:
/test all

@amygdala Can you try to rebase on the latest master to see if that fixes the problems?

The most recent test run had a zero output.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_examples/626/kubeflow-examples-presubmit/1166791544506683397/

amygdala:
(sorry about delay).  I think it might make the most sense to rebase the cookbook branch with master, then I will update my fork of the cookbook branch post-rebase.  I don't have edit privs on this repo to do the rebase, though -- would you be up for doing that?

jlewi:
@amygdala I made you a repo admin.

Would you mind opening an issue of kind process to track any changes you make to the branch just for future reference.

If its already documented somewhere you could just link to it; but otherwise in the issue it would be good to explain why cookbook branch is tracking master (i.e. being rebased periodically on master) as opposed to being more like a release branch.

jlewi:
Here are the prow logs 
[prow_job.log.txt](https://github.com/kubeflow/examples/files/3584664/prow_job.log.txt)

Its failing because ksonnet is having a segfault. This could be because the tests have been updated on master but not the branch in which case updating would help.

Alternatively if its safe to do so you could comment out one or more tests in the prow_config.yaml file on the cook book branch in prow_config.yaml
https://github.com/kubeflow/examples/blob/cookbook/prow_config.yaml

to disable that test.

amygdala:
Tracking issue: https://github.com/kubeflow/examples/issues/639

jlewi:
Test failures due to segfaults in ks are being tracked in kubeflow/testing#459

/lgtm
/approve

Issue #636, opened 2019-09-11
Updated example xgboost_synthetic to kubeflow.fairing


jinchihe:
Fairing will be a subpackage in kubeflow in https://github.com/kubeflow/fairing/issues/234, need to update the  example xgboost_synthetic to kubeflow.fairing once the PR merged https://github.com/kubeflow/fairing/pull/359

/assign @jinchihe 
/cc @jlewi

PR #637, opened 2019-09-12
update example to use kubeflow-fairing

jinchihe:
update example to use kubeflow-fairing as
```
import kubeflow.faring...
```
fixes: #636

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/637)


jinchihe:
@jlewi This is ready for go. please take a look. 
By the way I removed all cells outputs as that's more clear.

PR #647, opened 2019-09-27
Updated the azurepipeline example.

NikeNano:
I believe there is a small bug in the script, use tmp variable to solve the issue.


NikeNano:
@googlebot I signed it!

PR #658, opened 2019-10-12
pin the web-ui version of TF to 1.7-- same as training

amygdala:
Newer versions give an import error.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/658)

PR #664, opened 2019-10-18
Fixes Bugs for the mnist example

capt2101akash:
This PR has the bug fix for the mnist example for both the end to end flow and the ML pipeline flow.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/664)

@jlewi 
i) It specifies the TF version to 1.14.0 or 1.15.0 which is the latest stable version that supports TF serving instead of just specifying tensorflow.
ii) In the deploy service Dockerfile it changes the ksonnet version to 0.13.0 because in ksonnet version 0.12.0 there has been a bug reported i.e. Panic Error.
iii) Changes the docker image tag for serving which gives the Panic error while deploying.

### DIFF
--- a/mnist/serving/base/service.yaml
+++ b/mnist/serving/base/service.yaml
@@ -36,4 +36,4 @@ spec:
     targetPort: 8500
   selector:
     app: mnist
-  type: ClusterIP
+  type: LoadBalancer
--- a/mnist/serving/base/service.yaml
+++ b/mnist/serving/base/service.yaml
@@ -36,4 +36,4 @@ spec:
     targetPort: 8500
   selector:
     app: mnist
-  type: LoadBalancer
+  type: ClusterIP
--- a/mnist/web-ui/Dockerfile
+++ b/mnist/web-ui/Dockerfile
@@ -28,7 +28,7 @@ RUN pip --no-cache-dir install \
         h5py \
         ipykernel \
         numpy \
-        tensorflow \
+        tensorflow==1.14.0 \
         tensorflow-serving-api \
         flask \
         && \
--- a/pipelines/mnist-pipelines/mnist_pipeline.py
+++ b/pipelines/mnist-pipelines/mnist_pipeline.py
@@ -73,7 +73,7 @@ def mnist_pipeline(model_export_dir='gs://your-bucket/export',
 
   webui_args = [
           '--image', 'gcr.io/kubeflow-examples/mnist/web-ui:'
-                     'latest',
+                     'v20190304-v0.2-176-g15d997b',
           '--name', 'web-ui',
           '--container-port', '5000',
           '--service-port', '80',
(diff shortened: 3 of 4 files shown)


mnist/serving/base/service.yaml
```diff
@@ -36,4 +36,4 @@ spec:
     targetPort: 8500
   selector:
     app: mnist
-  type: ClusterIP
+  type: LoadBalancer
```
jlewi: This will expose the service on a public IP. We generally don't want to expose services on publicly accessible IP addresses by default because its unsafe. 

By default services should only be exposed through the Kubeflow ISTIO gateway so that it is secured by whatever auth is added to the gateway

Can we define a VirtualService to expose it through the ingress gateway?

Can we tell users to port-forward to the port rather than exposing it behind an LB?

pipelines/mnist-pipelines/mnist_pipeline.py
```diff
@@ -65,15 +65,15 @@ def mnist_pipeline(model_export_dir='gs://your-bucket/export',
...
+            '8dfa2d9f832e5291649e3bc9d58468da7c6ed92a',
       arguments=serve_args
   )
   serve.after(train)
 
 
   webui_args = [
           '--image', 'gcr.io/kubeflow-examples/mnist/web-ui:'
-                     'v20190304-v0.2-176-g15d997b-pipelines',
+                     'latest',
```
jlewi: Can we pin to a specific commit rather than using latest? 
The semantics of latest are to always point to the newest image which is brittle

jlewi:
@capt2101akash Thanks! Could you please update the PR description to explain what the issue is and how this fixes it?

mnist/serving/base/service.yaml
capt2101akash: Yeah, that makes sense. Will change the service type.

capt2101akash:
@jlewi changes made as per your request and updated the PR description.

capt2101akash:
@jlewi can you trigger the retest, I don't know why it failed?

jinchihe:
@capt2101akash I'm going to debug the issue. thanks.

mnist/web-ui/Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN pip --no-cache-dir install \
         h5py \
         ipykernel \
         numpy \
-        tensorflow \
+        tensorflow==1.14.0 \
```
jinchihe: @amygdala would list to pin the tensorflow to 1.7 in the PR https://github.com/kubeflow/examples/pull/658 since `from tensorflow.examples.tutorials.mnist import input_data` causing the problem, see more the #658 , could you please confirm if that works with 1.14.0 ? Thanks.

jlewi:
@capt2101akash Do you know why the test is failing?

jlewi:
Is this test failure: kubeflow/testing/issues/499

jinchihe:
@jlewi The test failed due to the kubectl version is too old for the cluster. I already work around the issue in the PR: https://github.com/kubeflow/examples/pull/668, could you please take a look? Thanks.

jlewi:
Looks like this is still blocked on #668

PR #669, opened 2019-10-24
Updates to the pipelines GH summarization lab to demonstrate component input/output

amygdala:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/669)


amygdala:
/hold
(not quite ready for review; first want to see if the tests are passing)

amygdala:
/assign @Ark-kun 
Alexey, here's that PR. I made some changes so that the components explicitly generate outputs to be consumed by other steps (rather than just constructing the arg paths from the workflow params as was done previously). There are a few other unrelated cleanup/refactoring changes. 
Thanks for taking a look.

amygdala:
Alexey, I just learned about the new `inputPath` from _your_ PR :), so I probably want to adjust this example to include its use. @Ark-kun

Ark-kun:
> Alexey, I just learned about the new `inputPath` from _your_ PR :), so I probably want to adjust this example to include its use. @Ark-kun
> Update: actually, I retract this-- I think it's okay as is. I'm only passing small strings, so `inputValue` is appropriate.

In future you might want to switch to it. Basically every "GCSpath" in a pipeline is a workaround for lack of support for artifact passing: The lack of "inputPath" and the fact that outputting big data caused failures before.

Then your pipeline will become portable (not GCP-bound) and able to run out of box (no `project='YOUR_PROJECT_HERE'` or `working_dir='YOUR_GCS_DIR_HERE'`).

github_issue_summarization/pipelines/example_pipelines/pipelines-notebook.ipynb
```diff
@@ -361,47 +357,42 @@
...
     "  log_dataset.after(copydata)\n",
-    "  train.after(copydata)\n",
     "  train.after(preproc)\n",
     "\n",
     "  log_model = metadata_log_op(\n",
     "    log_type=MODEL,\n",
     "    workspace_name=WORKSPACE_NAME,\n",
     "    run_name='{{workflow.name}}',\n",
-    "    model_uri='%s/%s/model_output' % (working_dir, '{{workflow.name}}')\n",
+    "    model_uri=copydata.outputs['copy_output_path']\n",
```
Ark-kun: Should if be
```python
model_uri=train.outputs['train_output_path'],
```
?

github_issue_summarization/pipelines/example_pipelines/pipelines-notebook.ipynb
```diff
@@ -361,47 +357,42 @@
...
-    "  train.set_gpu_limit(4)\n",
+    "  train.set_gpu_limit(1)\n",
     "  train.set_memory_limit('48G')\n",
     "\n",
-    "  with dsl.Condition(train.output == 'true'):\n",
+    "  with dsl.Condition(train.outputs['launch_server'] == 'true'):\n",
     "    webapp = dsl.ContainerOp(\n",
     "        name='webapp',\n",
     "        image='gcr.io/google-samples/ml-pipeline-webapp-launcher:v2ap',\n",
     "        arguments=[\"--model_name\", 'ghsumm-%s' % ('{{workflow.name}}',),\n",
```
Ark-kun: '{{workflow.name}}' should ideally be replaced with `kfp.dsl.RUN_ID_PLACEHOLDER` (so that we do not expose Argo). (`RUN_ID_PLACEHOLDER` == `{{workflow.uid}}`, not `{{workflow.name}}`, but there should not be a big difference).

github_issue_summarization/pipelines/example_pipelines/pipelines-notebook.ipynb
amygdala: yes -- thanks, good catch

amygdala:
> > Alexey, I just learned about the new `inputPath` from _your_ PR :), so I probably want to adjust this example to include its use. @Ark-kun
> > Update: actually, I retract this-- I think it's okay as is. I'm only passing small strings, so `inputValue` is appropriate.
> 
> In future you might want to switch to it. Basically every "GCSpath" in a pipeline is a workaround for lack of support for artifact passing: The lack of "inputPath" and the fact that outputting big data caused failures before.
> 
> Then your pipeline will become portable (not GCP-bound) and able to run out of box (no `project='YOUR_PROJECT_HERE'` or `working_dir='YOUR_GCS_DIR_HERE'`).

That's a good idea for the next pass.

github_issue_summarization/pipelines/example_pipelines/pipelines-notebook.ipynb
amygdala: I went ahead and made this change as well.

amygdala:
I fixed that typo re: passing the wrong arg, and also changed the pipeline to use kfp.dsl.RUN_ID_PLACEHOLDER.  PTAL again when you get a chance.  @Ark-kun

PR #719, opened 2020-02-06
Fix namespace test runs

gabrielwen:
New namespace should be kubeflow-kf-ci-v1-user

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/719)


gabrielwen:
@jlewi this PR is ready and should fix the issue.

jlewi:
Thanks. 

/lgtm
/approve

Issue #713, opened 2020-01-24
mnist example: change to deploy tfjob into user profile namespace


amygdala:
Looks like the mnist example is broken; looks like the tfjob needs to be deployed into a user profile namespace now. The quick workaround is to just direct the output of 'kustomize build' to a yaml file rather than piping directly to kubectl; then editing that file to change the namespace, then applying it. The more principled fix would be to change the set of kustomize commands.

(for googlers, see [this discussion](https://groups.google.com/a/google.com/g/google-kubeflow-support/c/ZQixo8knLiA/m/QI9JijprEQAJ) for background)

I filed this bug as well: https://github.com/kubeflow/website/issues/1554 since the docs use this code for an [e2e tutorial](https://www.kubeflow.org/docs/gke/gcp-e2e/).

amygdala:
probably we should bump the priority higher, since this code is used for the main docs 'getting started' example.

amygdala:
@sarahmaddox @jlewi This is the one that really needs the priority bumped in my opinion.  I was thinking to do a quick fix by having them redirect the kustomize build output to a yaml, then having them edit the yaml to change the 'kubeflow' namespace to their user one.

amygdala:
Pulled in https://github.com/kubeflow/examples/pull/720 , but I like Jeremy's suggestion there about how to make the fix a bit nicer.  I'll make the same change in the docs e2e page too, which uses this example.

amygdala:
Update: See also https://github.com/kubeflow/website/issues/1554 . Sounds like @jlewi will tackle a notebook-based version.

PR #720, opened 2020-02-07
mnist example namespace fix

amygdala:
This is necessary b/c the kubeflow namespace, by default, doesn't have sufficient permissions.
Addresses https://github.com/kubeflow/examples/issues/713.

This is a quick fix, just so it continues to work. When there's time, we can figure out how to address it via kustomize config instead. 

cc @DanSanche  -- LMK you'd rather approach this differently.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/720)

### DIFF
--- a/mnist/README.md
+++ b/mnist/README.md
@@ -229,20 +229,32 @@ kustomize edit add configmap mnist-map-training --from-literal=modelDir=gs://${B
 kustomize edit add configmap mnist-map-training --from-literal=exportDir=gs://${BUCKET}/${MODEL_PATH}/export
 ```
 
-
-You can now submit the job
+Build a yaml file for the `TFJob` specification based on your kustomize config:
 
 ```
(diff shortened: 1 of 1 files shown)


jlewi:
/hold
/lgtm
/approve

@amygdala I think this is a good solution. Two other things we could do that might decrease likelihood of people running in kubeflow namespace

1. Remove namespace from train.yaml
    * If users didn't explicitly set it; then it would default to whatever the kubectl context is

2. Add a namespace which doesn't exist e.g "your-profile-namespace" so if users don't change it they will get an error because the namespace doesn't exist.

I'm fine what you have so feel free to just remove the hold.

amygdala:
Thanks, Jeremy.  Let me pull this one in, then do another iteration to address your suggestions (I'll probably go with #2).

PR #723, opened 2020-02-11
Create a notebook for mnist E2E on GCP

jlewi:
A notebook to run the mnist E2E example on GCP.

This fixes a number of issues with the example
* Use ISTIO instead of Ambassador to add reverse proxy routes
* The training job needs to be updated to run in a profile created namespace in order to have the required service accounts
     * See kubeflow/examples#713
     * Running inside a notebook running on Kubeflow should ensure user
       is running inside an appropriately setup namespace
* With ISTIO the default RBAC rules prevent the web UI from sending requests to the model server
     * A short term fix was to not include the ISTIO side car
     * In the future we can add an appropriate ISTIO rbac policy
     
* Using a notebook allows us to eliminate the use of kustomize
  * This resolves kubeflow/examples#713 which required people to use
    and old version of kustomize

  * Rather than using kustomize we can use python f style strings to
    write the YAML specs and then easily substitute in user specific values

  * This should be more informative; it avoids introducing kustomize and
    users can see the resource specs.

* I've opted to make the notebook GCP specific. I think its less confusing
  to users to have separate notebooks focused on specific platforms rather
  than having one notebook with a lot of caveats about what to do under
  different conditions

* I've deleted the kustomize overlays for GCS since we don't want users to
  use them anymore

* I used fairing and kaniko to eliminate the use of docker to build the images
  so that everything can run from a notebook running inside the cluster.

* k8s_utils.py has some reusable functions to add some details from users
  (e.g. low level calls to K8s APIs.)

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/723)


jlewi:
/assign @amygdala 
/assign @jinchihe 

@amygdala I think this is ready for review. There are some lint issues but I think that might be blocked by kubeflow/testing#560

amygdala:
this sounds great, will take a look today

jlewi:
kubeflow/examples#728 removes the failing lint test. Once that is merged I will rebase this and the tests should pass.

mnist/training/train.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: kubeflow.org/v1
+kind: TFJob
+metadata:
+  name: $(trainingName)
+  namespace: kubeflow
```
amygdala: haven't tested yet -- but would this have the same issue of needing to swap in the user profile namespace?

amygdala:
Some testing notes, that I will update as I run through the notebook:

- It looks like 'retrying' needs to be installed too. E.g. adding a cell with `!pip install --user retrying` worked for me
- I had to restart the kernel after the running the `notebook_setup`,  to find `k8s_util`

mnist/mnist_gcp.ipynb
```diff
@@ -0,0 +1,895 @@
...
+      "Checkout kubeflow/tf-operator @9238906\n",
+      "Configure docker credentials\n",
+      "Adding /home/jovyan/git_tf-operator/sdk/python to python path\n"
+     ]
+    }
+   ],
+   "source": [
+    "import notebook_setup\n",
+    "reload(notebook_setup)\n",
+    "notebook_setup.notebook_setup()"
```
amygdala: I think a kernel restart is needed at this point.  Also, I needed to install 'retrying' too, do the kernel restart  (I added a cell with `!pip install --user retrying`)

mnist/mnist_gcp.ipynb
```diff
@@ -0,0 +1,895 @@
...
+    "from kubernetes.client import rest as k8s_rest\n",
+    "from kubeflow import fairing   \n",
+    "from kubeflow.fairing import utils as fairing_utils\n",
+    "from kubeflow.fairing.builders import append\n",
+    "from kubeflow.fairing.deployers import job\n",
+    "from kubeflow.fairing.preprocessors import base as base_preprocessor\n",
+    "\n",
+    "# Setting up google container repositories (GCR) for storing output containers\n",
+    "# You can use any docker container registry istead of GCR\n",
+    "GCP_PROJECT = fairing.cloud.gcp.guess_project_name()\n",
```
amygdala: Here, I think it would be useful to print out the guessed project name, in case anything did not work as expected. e.g. just `print('using GCP project {}'.format(GCP_PROJECT))`

mnist/mnist_gcp.ipynb
```diff
@@ -0,0 +1,895 @@
...
+    "DOCKER_REGISTRY = 'gcr.io/{}/fairing-job'.format(GCP_PROJECT)\n",
+    "namespace = fairing_utils.get_current_k8s_namespace()"
```
amygdala: same suggestion re: printing out the namespace

mnist/mnist_gcp.ipynb
```diff
@@ -0,0 +1,895 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Create the training job\n",
+    "\n",
+    "* We could write the spec to a YAML file and then do `kubectl apply -f {FILE}`\n",
+    "* Since we are running in jupyter we will use the Kubernetes python client"
```
amygdala: It might be helpful to also add a comment that re: 'we are deploying this job into your user profile namespace'

mnist/mnist_gcp.ipynb
```diff
@@ -0,0 +1,895 @@
...
+    "tb_specs = [tb_deploy, tb_service, tb_virtual_service]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "k8s_util.apply_k8s_specs(tb_specs, k8s_util.K8S_CREATE_OR_REPLACE)"
```
amygdala: Here I'm getting this error.  I actually have a bug filed for a similar thing, as the Github issue summ pipeline ex hits a related issue (let me find the bug #).  For that codelab I have them run a `create clusterrolebinding sa-admin...` command.
```
ApiException: (403)
Reason: Forbidden
HTTP response headers: HTTPHeaderDict({'Audit-Id': 'f3f47933-7471-4516-8b11-66a1e1b1e414', 'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'Date': 'Wed, 12 Feb 2020 16:32:10 GMT', 'Content-Length': '405'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"virtualservices.networking.istio.io is forbidden: User \"system:serviceaccount:kubeflow-amyu:default-editor\" cannot create resource \"virtualservices\" in API group \"networking.istio.io\" in the namespace \"kubeflow-amyu\"","reason":"Forbidden","details":{"group":"networking.istio.io","kind":"virtualservices"},"code":403}
```
Running `kubectl create clusterrolebinding sa-admin --clusterrole=cluster-admin   --serviceaccount=kubeflow-amyu:default-editor` fixed the error.

mnist/mnist_gcp.ipynb
```diff
@@ -0,0 +1,895 @@
...
+    "## Wait For the Training Job to finish"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tf_job = tf_job_client.wait_for_condition(train_name, expected_condition=[\"Succeeded\", \"Failed\"], namespace=namespace)"
```
amygdala: could this add a `print(tf_job)` or similar so that they can see if the job succeeded or failed?

amygdala:
This is really great!  Added comments on a few minor glitches.

A general comment: people will also want to know how to view the output of the training logs (particularly for when they swap in their own model) — could we add a section on that?

mnist/training/train.yaml
jlewi: Woops this should have been removed. Originally I thought I would have YAML files that I would load in the notebook. But I decided it was more illustrative to inline the YAML in the notebook and use f style strings for substitution. I will remove this.

mnist/mnist_gcp.ipynb
jlewi: I'm like to try to find a way to avoid the kernel restart. Because I'd like to just be able to tell people to do "Run All."

I believe if we update the path the imports should be locatable immediately. If you've already loaded the module a reload might be needed.

mnist/mnist_gcp.ipynb
jlewi: Looks like kubeflow/manifests/issues/891

jlewi:
Thanks @amygdala I added a section to print out links to stackdriver where the logs can be viewed.

mnist/mnist_gcp.ipynb
amygdala: Hmm, it looked from the output like the paths were updated when I ran `notebook_setup()`, but I still had to restart.  Maybe we should see if someone else can repro. (@jinchihe , if you run through it too, let us know if this part worked for you).

mnist/mnist_gcp.ipynb
jlewi: Done.

mnist/mnist_gcp.ipynb
jlewi: Done.

mnist/mnist_gcp.ipynb
jlewi: Done.

mnist/mnist_gcp.ipynb
jlewi: Done.

mnist/mnist_gcp.ipynb
jlewi: I tried the latest version in a new notebook
https://github.com/jlewi/examples/commit/5b1262fe6255776d03dbded7d4d500a04508491f

The imports worked fine.
I have added reload statements since the original verison.

mnist/mnist_gcp.ipynb
jlewi: I verified that kubeflow/manifests#891 is now fixed at
https://github.com/kubeflow/manifests/tree/096f312

jlewi:
@amygdala This is ready for another look

* I added the print statements you suggested
* The RBAC permissions for virtual services kubeflow/manifests/issues/891; has been fixed in the latest version on the v1.0 branch
* I added a section about fetching the logs from stackdriver
* I ran the notebook E2E on a fresh notebook server and it ran successfully no import errors
  * I suspect the issue might be if you have already issued an import you need to issue a reload to
     pick up any changes in the module
  * Further complicating matters is that the kubeflow python imports share the top level package kubeflow
    * So I think after installing fairing, and tfjob sdk we need to do `reload(kubeflow)`; I added a couple
      of those statements and things seem to be working.

amygdala:
Great.  I may not get a chance to look at it until Sat. as I am at a conf the next couple of days.

jlewi:
Thanks @amygdala 

@jinchihe any chance you could review this? I think the sooner we can get this merged will be beneficial to the folks working on versions of this notebook for other platforms.

jinchihe:
It's amazing!!! @jlewi 
Great for user experience and exciting to use `kubeflow-tfjob` SDK. 
I'm going to enhance the example for on-premise cluster as tne notebook way, now I'm thinking maybe better to use KFServing to deploy serving, will try that on on-premise cluster.

/lgtm
/approve

PR #736, opened 2020-02-19
deprecating gis e2e example until it is fixed.

amygdala:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/736)


amygdala:
I tried to figure out how to change the font color in order to e.g make a red notice (I thought I could do it via raw html), but I don't believe the github markdown processing actually supports that.

Issue #725, opened 2020-02-11
Update mnist E2E example for on prem vanilla KF


jlewi:
There are a number of issues with the mnist example that need to be addressed for KF 1.0
https://github.com/kubeflow/examples/tree/master/mnist

1. The example is using Ambassador not ISTIO for routes to the UI and model
1. The training job needs to be updated to run in a profile created namespace in order to have
the required service accounts

     * See kubeflow/examples#713
1. With ISTIO the default RBAC rules will prevent the web UI from sending requests to the model server
    * We either need to create an appropriate ISTIO rbac authorization policy or just remove the ISTIO sidecar
1. There is a requirement to use an old version of kustomize kubeflow/examples#713

     * I think we should probably just remove kustomize
     * kustomize might be increasing code reuse for different versions (e.g. GCP vs. AWS) but it hurts the readability IMO
     * kustomize gets in the way of the YAML spec which is what we want to show/teach users just getting started with Kubeflow

In addition the vanilla/on prem/non-cloud specific distribution of Kubeflow has an added question of what to do about storage? The current solution uses a PV which works but isn't compatible with distributed training.

On GCP the solution we've opted for is to create a notebook that runs the entire example

- Draft pull request kubeflow/examples#723

A notebook has the following advantages

- The user can just do "Run All" to run all the cells; eliminates a lot of copying and pasting
- We can more easily add E2E testing to verify the notebook runs E2E 
- We can use rich markdown to provide detailed explanations of what's happening
- We can eliminate kustomize and just use Python f style strings
- This allows us to show users the raw YAML while easily updating it programmatically for their
  specific use case
- Any complex or shared logic can be "hidden" in a python library that is imported

The purpose of this bug is to figure out what we want to do about mnist for the  on prem distribution (e.g. kfctl_k8s_istio.yaml). It would be nice to have a similar mnist E2E example for all Kubeflow distributions.

My recommendation would be that we have separate notebooks for different Kubeflow distributions/clouds. This will lead to some code duplication but I think it provides a better learning experience. If we have one uber notebook I think we will end up with a lot of "If you are on Cloud X then do Y" and I think that hurts users.

In particular, if this version of the  example uses a PVC the YAML specs start to look different from Cloud solutions using object storage. I think putting all the variants into one notebook will be confusing.

/cc @krishnadurai @yanniszark

jlewi:
Could we reuse the minio instance that KFP ships?

swiftdiaries:
[...]

PR #737, opened 2020-02-19
Mnist vanilla k8s

swiftdiaries:
Fixes: https://github.com/kubeflow/examples/issues/725

/cc @johnugeorge @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/737)


mnist/mnist_vanilla_k8s.ipynb
```diff
@@ -0,0 +1,909 @@
+{
```
jlewi: <p>Delete credits; not neccessary.</p><p> </p><p>In the overview can you describe the storage solution e.g minio</p><p> </p><p>Also it looks like this depends on DockerHub for the registry. Could you call that out as well?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/kubeflow/examples/pull/737/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

jlewi: <p>Can we automate the configmap creation with python code so they just have to execute the cell to create the configmap?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/kubeflow/examples/pull/737/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

jlewi: <p>Why do you have input statements? Couldn't the user just set these python variables?</p><p> </p><p>How about moving DOCKER_REGISTRY further up in the notebook? Ideally a user would just have to set one or two values in the first cell and then click run all to have the notebook run</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/kubeflow/examples/pull/737/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

jlewi: <p>nit: I think its cleaner to use the python client libraries to get these values rather than shelling out to kubectl.</p><p> </p><p>More importantly is it possible to print a warning but keep going in the event the user lacks permission? </p><p> </p><p>Ideally, a user would just be able to click run all and it would run top to bottom. Since printing out the URLs isn't critical it would be best if it printed out a warning but otherwise kept going.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/kubeflow/examples/pull/737/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

swiftdiaries: Deleted

swiftdiaries: <p>Added instructions on how to add docker registry and why it's used</p>

swiftdiaries: <p>We haven't made a standardized way to authenicate to DockerHub as part of the vanilla-k8s deployment.</p><p>I feel like it'd be better to addressed at creation or have a standardized way like <code>gcloud configure docker </code> </p><p>So as of now, it's okay for it to be separate?</p>

swiftdiaries: <p>Changed it out to get this from the k8s python client</p>

swiftdiaries: <p>Changed it out to use the k8s python client library and warn if user is not authenticated.</p><p>Added instructions on fetching the URLs at the end of the notebook in case the user wants to do it manually.</p>

swiftdiaries: <p>Moved DOCKER_REGISTRY to the beginning of the notebook and added instructions to change it out</p>

swiftdiaries:
@jlewi This is ready for another look

PR #741, opened 2020-02-20
Fix the mnist_gcp_test.py

jlewi:
Fix the mnist_gcp_test.py

* The job spec was invalid; we were missing container name

* There were a bunch of other issues as well.

* Pull in the changes from xgboost_synthetic to upload an HTML version
  of the notebook output to GCS.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/741)

### DIFF
--- a/py/kubeflow/examples/notebook_tests/execute_notebook.py
+++ b/py/kubeflow/examples/notebook_tests/execute_notebook.py
@@ -1,4 +1,4 @@
-import fire
+import argparse
 import tempfile
 import logging
 import os
@@ -6,10 +6,13 @@ import subprocess
 
 logger = logging.getLogger(__name__)
 
+from google.cloud import storage
+from kubeflow.testing import util
+
 def prepare_env():
   subprocess.check_call(["pip3", "install", "-U", "papermill"])
-  subprocess.check_call(["pip3", "install", "-r", "../requirements.txt"])
-
+  subprocess.check_call(["pip3", "install", "-U", "nbconvert"])
+  subprocess.check_call(["pip3", "install", "-U", "nbformat"])
 
 def execute_notebook(notebook_path, parameters=None):
   import papermill #pylint: disable=import-error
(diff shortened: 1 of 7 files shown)


jlewi:
Success: Here's a link to the uploaded notebook
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubeflow_examples/741/kubeflow-examples-presubmit/1230637341891104769/artifacts/notebooks-v1/mnist-233822-fe3.html

jlewi:
/assign @gabrielwen 

This is ready for review.

py/kubeflow/examples/notebook_tests/execute_notebook.py
```diff
@@ -21,13 +24,32 @@ def execute_notebook(notebook_path, parameters=None):
...
-      assert False, "Unable to find from output: " + expected_message
+
+  with open(output_path, "r") as hf:
+    actual_output = hf.read()
+
+  nb = nbformat.reads(actual_output, as_version=4)
+  html_exporter = nbconvert.HTMLExporter()
+  (html_output, _) = html_exporter.from_notebook_node(nb)
+  gcs_path = os.getenv("OUTPUT_GCS")
+  _upload_notebook_html(html_output, gcs_path)
```
gabrielwen: I didn't anywhere to check whether the notebook succeeds or not?

jlewi: Checking for expected output from specific cells is brittle. My expectation would be that if a notebook successfully runs end to end then it is working. If there is a problem then the cell should raise an exception which would terminate execution.

I verified that is the case.
I added a cell that would raise an exception and pushed that as a commit. here's the test run
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_examples/741/kubeflow-examples-presubmit/1230975731350114304/

You can see that the test failed.

Here's what happens

1. If a notebook cell raises an exception this exception gets propogated through papermill.execute_notebook
    * Here are the logs from [execute_notebook](https://pantheon.corp.google.com/logs/viewer?project=kubeflow-ci-deployment&folder&organizationId&minLogLevel=0&expandAll=false&timestamp=2020-02-20T21:18:57.486000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeStart=2020-02-21T16:06:33.041Z&dateRangeEnd=2020-02-21T22:06:33.041Z&interval=PT6H&resource=k8s_container%2Fcluster_name%2Fkf-master-0218-beb%2Fnamespace_name%2Fkubeflow-kf-ci-v1-user%2Fcontainer_name%2Fexecuting-notebooks&advancedFilter=resource.type%3D%22k8s_container%22%0Aresource.labels.namespace_name%3D%22kubeflow-kf-ci-v1-user%22%0Aresource.labels.container_name%3D%22executing-notebooks%22%0Alabels.%22k8s-pod%2Fjob-name%22%3D%22mnist-220258-e9d%22&scrollTimestamp=2020-02-21T22:05:54.291166423Z)

1. The exception causes execute_notebook.py to fail (exit with non-zero exit code)
1.  As a result the K8s job running execute_notebook.py gets marked as failed
1. As a result the pytest raises an exception

gabrielwen: ok, got it.  so it's when an exception was thrown.

PR #742, opened 2020-02-20
Update mnist example for on-prem cluster

jinchihe:
The PR is going to update mnist example for on-prem cluster.
Related with #725 

I noticed @swiftdiaries also created one PR to update the example for vanilla k8s cluster, and @Jeffwan created one for AWS. That looks great. 

Here I have some personal thoughts, so I create new one to show, let's have more discussion, maybe we need to merge or take one of this. 
My thoughts:
1. We should take kubeflow Components as more as possible in the example to show the Kubeflow key components. In the PR, I'm using Kubeflow Fairing to build docker image and launch TFJob to training model.
2. I used `set_preprocessor`, `set_builder` and `set_deployer`, that's more common for Fairing usage.
```
fairing.config.set_preprocessor('python', command=command, path_prefix="/app", output_map=output_map)
fairing.config.set_builder(name='docker', registry=DOCKER_REGISTRY, base_image="",
                           image_name="mnist", dockerfile_path="Dockerfile")
fairing.config.set_deployer(name='tfjob', namespace=my_namespace, stream_log=False,
                            worker_count=num_workers, ps_count=num_ps,
                            pod_spec_mutators = [mounting_pvc(pvc_name=pvc_name, pvc_mount_path='/mnt')])
```
3. And then using Fairing to deploy a InferenceService (KFServing has been integrated with Fairing)
```
from kubeflow.fairing.deployers.kfserving.kfserving import KFServing
isvc = KFServing('tensorflow', default_storage_uri='pvc://' + pvc_name + '/export', namespace=my_namespace)
isvc.deploy(isvc.generate_isvc())
```
4. I still take NFS PV/PVC to store exported model, for now, the Fairing, KFServing, TFJob support PVC and for local cluster, the shared PVC is very popular and easy.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/742)


jlewi:
I think there's a question about how much we want to highlight KFServing in the example. One of the goals of the mnist example is to highlight the core CUJ for 1.0. Since KFServing isn't going 1.0 there's some open question about whether we should feature it or not.

In terms of highlighting fairing patterns. One of the reasons in my original notebook that I didn't use preprocessort and builder is because I think some of those patterns might be worth rethinking. 

In this particular case I don't think the preprocessor is doing much since we aren't extracting code from the notebook. So IME fairing is mostly adding magic that I usually end up reverse engineering.

I filed an issue  kubeflow/fairing#454.I think we might want to look at using Tekton tasks to run docker builds in cluster. 

* That would allow us to swap in different builders (kaniko, buildkit)
* Provide a more consistent experience; i.e. define a YAMl spec and then just apply it (like we do for TFJob) (kubeflow/fairing#453)
* Leverage Tekton to easily pull context form different places (git, GCS, S3, etc...).

I think the functionality that fairing would provide would be

* Python utilities for creating a context to work with tekton
* Python utilities for applying K8s manifests: kubeflow/fairing#453

So I think fairing is probably due for an overhaul in terms of goals, API, UX, etc... I'm kind of hoping you and/or IBM will take this on .

So given I expect fairing to change pretty substantially; I personally prefer to highlight a lower level, K8s native experience that is unlikely to be impacted by the refactor.

That said, I'm fine with having an example that highlights more of how fairing works today. Mostly just want to explain my thought process.

/cc @animeshsingh

marcindulak:
Is using docker fairing builder the best choice? When launching jupyter servers from kubeflow the choice of maintained images is limited, the default offered ones include `gcr.io/kubeflow-images-public/tensorflow-1.14.0-notebook-cpu:v0.7.0` + gpu version, and the corresponding tensorflow 2 ones. Those images don't have docker installed. I've been therefore trying to use append builder,
but encountered https://github.com/kubeflow/fairing/issues/461 and https://github.com/kubeflow/fairing/issues/460. In which direction the instructions for fairing builders will be going?

jinchihe:
Thanks @jlewi 

If so I think we can check this in the Kubeflow Fairing repo to be a E2E user case for `kubeflow-fairing`. There are someones want to use KFServing with PVC to deploy service because I noticed there are some issues logged about this.  And the #737 has been checked on, that's good enough for on-prem cluster. 

@marcindulak I will take a look you problem in Fairing repo. Thanks.

Issue #738, opened 2020-02-19
[codelab-intro-mnist-feedback]: Dockerfile.model in examples/mnist needs to be fixed


JIElite:
Issue: Dockerfile.model in examples/mnist needs to be fixed
Platform: Google Cloud Platform with Cloud Shell

According to "[Introduction to Kubeflow on Google Kubernetes Engine](https://codelabs.developers.google.com/codelabs/kubeflow-introduction/index.html#2)", we can run the following command to run docker container:
`docker run -it $IMAGE_PATH`

However, I ran into this error message at 21:40(PST) on 2020/2/18:
> docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"/usr/bin/python\": stat /u
sr/bin/python: no such file or directory": unknown.
ERRO[0000] error waiting for container: context canceled

The content in Docker.model is:
```
#This container contains your model and any helper scripts specific to your model.
# When building the image inside mnist.ipynb the base docker image will be overwritten
FROM tensorflow/tensorflow:1.15.2-py3

ADD model.py /opt/model.py
RUN chmod +x /opt/model.py

ENTRYPOINT ["/usr/bin/python"]
CMD ["/opt/model.py"]
```

After changing the entrypoint to `ENTRYPOINT ["/usr/local/bin/python"]`, the `docker run` command properly.

Is this dockerfile still in progress? or just leave something that no mentioned in the document.

Best regards,
Eli

millerdz:
Submitted PR #744 to fix this issue.

PR #744, opened 2020-02-20
Update ENTRYPOINT to use /usr/local/bin/python in Dockerfile.model

millerdz:
Base image `FROM tensorflow/tensorflow:1.15.2-py3` uses python3 and therefore the python binary location is `/usr/bin/python3`. However, [tensorflow base image creates a symlink](https://github.com/tensorflow/tensorflow/blob/e5bf8de410005de06a7ff5393fafdf832ef1d4ad/tensorflow/tools/dockerfiles/dockerfiles/cpu.Dockerfile#L45) to the current python binary as `/usr/local/bin/python` regardless if that is python version 2 or version 3, so that binary location should be used in the *ENTRYPOINT* of the `Dockerfile.model` instead of `/usr/bin/python` which is customary for Python v2.x installations. By using `/usr/bin/python` as the `ENTRYPOINT` with the tensorflow:*-py3 Docker images, users of this example will receive the error below as reported in #738:

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: "/usr/bin/python": stat /usr/bin/python: no such file or directory": unknown.
ERRO[0000] error waiting for container: context canceled
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/744)


mnist/Dockerfile.model
```diff
@@ -5,5 +5,5 @@ FROM tensorflow/tensorflow:1.15.2-py3
 ADD model.py /opt/model.py
 RUN chmod +x /opt/model.py
 
-ENTRYPOINT ["/usr/bin/python"]
+ENTRYPOINT ["/usr/bin/python3"]
```
jlewi: Should we change this to just be "python"

It looks like that is pointing to the correct version.

```
root@jlewi-tf2:/# which python
/usr/local/bin/python
root@jlewi-tf2:/# python --version
Python 3.6.9
```

It looks like that is what the notebook mnist_gcp.ipynb is doing which is why it doesn't hit the problem.

millerdz: Double-checked and the [tensorflow base image actually creates a symlink](https://github.com/tensorflow/tensorflow/blob/e5bf8de410005de06a7ff5393fafdf832ef1d4ad/tensorflow/tools/dockerfiles/dockerfiles/cpu.Dockerfile#L45) to the current python binary as `/usr/local/bin/python` regardless if that is python version 2 or version 3. So just changing this from `/usr/bin/python` to `/usr/local/bin/python` should be a more generic fix. I've updated the commit to clarify this as well as changing the python path in the ENTRYPOINT. 

Let me know if this is more suitable. 😄

millerdz:
@googlebot I signed it!

jlewi:
Looks good; but please rebase to pull in the latest changes which include a presubmit test for the mnist example. It looks like the test is currently is failing
https://github.com/kubeflow/examples/issues/750

So we may need to get that fixed before we merge this PR.

millerdz:
Merged w/ master.

jlewi:
/lgtm
/approve

Thanks

PR #745, opened 2020-02-21
Remove kustomize from mnist example.

jlewi:
* The mnist E2E guide has been updated to use notebooks and get rid
  of kustomize

* We have notebooks for AWS, GCP, and Vanilla K8s.

* As such we no longer need the old, outdated kustomization files or
  Docker containers anymore

  * The notebooks handle parameterizing the K8s resources using Python
    f style string.

* Update the README to remove the old instructions.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/745)


Jeffwan:
Look good to me

PR #747, opened 2020-02-23
Update mnist_gcp.ipynb - increase training step value

amygdala:
Increasing the train_steps value allows training progress info to show up in the logs.  (Apparently 200 steps was too small for that).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/747)

### DIFF
--- a/mnist/mnist_gcp.ipynb
+++ b/mnist/mnist_gcp.ipynb
@@ -341,7 +341,7 @@
     "num_workers = 2\n",
     "model_dir = f\"gs://{bucket}/mnist\"\n",
     "export_path = f\"gs://{bucket}/mnist/export\" \n",
-    "train_steps = 200\n",
+    "train_steps = 2000\n",
     "batch_size = 100\n",
     "learning_rate = .01\n",
     "image = cluster_builder.image_tag\n",


amygdala:
It doesn't seem that the papermill-related presubmit error is related to my mini change.

jlewi:
https://github.com/kubeflow/examples/issues/750 is tracking the notebook test failures.

jlewi:
@amygdala how much longer does training take if we increase from 200 to 2000 steps?

I'd like the notebook to run e2e pretty quickly say <=10 minutes. This is useful for testing (at least until we parameterize it better).

But also I think its valuable if people can just do "Run all" and have the notebook complete fairly quickly even if that means training a poor quality model.

Could we add a note instead along the lines of "To train a high quality model increase the number of steps to X? and it will take this long?"

Perhaps we could produce a graph from tensorboard that shows accuracy and training time as a function of steps and then include that in the markdown so that we can provide good guidance on the tradeoff between steps, time, and accuracy?

amygdala:
It actually didn't seem to take appreciably longer (b/c the model is so simple, it trains very quickly) but I will leave it up to you.  If you'd rather leave it like it is, that's fine.

jlewi:
@amygdala Did you want to rebase this and fix or should we just close this PR?

amygdala:
let's just close it, I don't think it's that important.

PR #752, opened 2020-02-24
[WIP] Tekton testing

gabrielwen:
DO NOT SUBMIT
testing codes for tekton pipelines

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/examples/752)


jlewi:
Thanks @gabrielwen 

A couple high level comments

* I think we want to start building up a catalog of tasks ala - https://github.com/tektoncd/catalog
* Should we create a new repository for this or just make it a directory in one of our existing repositories e.g. kubeflow/testing?
* I think we want to avoid using PVC to manage data this is an operational pain point since the NFS share fills up. I would suggest having a single task with multiple steps so we can pass data via the tekton workspace which is a pod volume.

gabrielwen:
@jlewi 
any pros of creating a new repo?  I think we can treat kubeflow/testing as our catalog?

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: pytest
```
jlewi: pytest is too generic; this test is testing a notebook so maybe something like "test-notebook"

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
...
+  annotations:
+    sidecar.istio.io/inject: "false"
+spec:
+  inputs:
+    params:
+    - name: test-worker-img
+      type: string
+      description: Test worker image.
+      default: gcr.io/kubeflow-ci/test-worker:latest
+    - name: src-dir
```
jlewi: Why is src-dir a parameter? My expectation would be that all source we provided using tekton resources corresponding to git resources. So you you would then reference the location as 
/workspace/$(inputs.resources.<repo name>.name)

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
...
+      description:
+        Job type of the test, from prow.
+      default: manual
+    - name: test_hostname
+      type: string
+      description:
+        host name of the test, from prow.
+      default: manual-testing
+  steps:
+  - name: git-checkout
```
jlewi: Why do you have a git checkout step? Why not just use Tekton git resources and let Tekton checkout the source for you?

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
...
+    image: $(inputs.params.test-worker-img)
+    command: ["/usr/local/bin/checkout_repos.sh"]
+    args:
+    - --repos=$(inputs.params.repos)
+    - --src_dir=$(inputs.params.src-dir)
+    volumeMounts:
+    - mountPath: $(inputs.params.src-dir)
+      name: testing-dir
+  - name: get-credentials
```
jlewi: Why

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
...
+    image: $(inputs.params.test-worker-img)
+    command: ["$(inputs.params.get-cred-cmd)"]
```
jlewi: Why is the command a parameter? Why can't we just hard code it?

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
...
+      value: $(inputs.params.gcp-sa-dir)/kubflow-releasing.json
+    volumeMounts:
+    - mountPath: $(inputs.params.src-dir)
+      name: testing-dir
+    - mountPath: $(inputs.params.gcp-sa-dir)
+      name: kubeflow-releasing-credential
+  - name: run-pytest
+    image: $(inputs.params.test-worker-img)
+    script: |
+      #!/usr/bin/env bash
```
jlewi: Why do we need a custom script? Why can't we just run papermill?

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
...
+      value: $(inputs.params.test_job_type)
+    - name: HOSTNAME
+      value: $(inputs.params.test_hostname)
+    volumeMounts:
+    - mountPath: $(inputs.params.src-dir)
+      name: testing-dir
+    - mountPath: $(inputs.params.gcp-sa-dir)
+      name: kubeflow-releasing-credential
+  volumes:
+  - name: testing-dir
```
jlewi: Do we need this? Doesn't tekton already create an empty dir /workspace that we can use to pass data back and forth?

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
...
+    params:
+    - name: test-worker-img
+      type: string
+      description: Test worker image.
+      default: gcr.io/kubeflow-ci/test-worker:latest
+    - name: src-dir
+      type: string
+      description: The path for source codes.
+      default: /mnt/testdir/src
+    - name: repos
```
jlewi: Can we use tekton resources to check out the repositories we need?

tekton_configs/pytest-task.yaml
```diff
@@ -0,0 +1,110 @@
...
+      type: string
+      description:
+        Repositories we need to checkout.
+    - name: python-path
+      type: string
+      description:
+        PYTHONPATH env.
+    - name: gcp-sa-dir
```
jlewi: Does this need to a parameter? I think we should try to use as few parameters as possible. 

I think for now you could hard code the secret inside the task to match what's in the cluster. The secret will be a property of the cluster not the individual task run so I don't think it should be a parameter.

i.e. the secret should be configured by the person installing the Task in the cluster not by users consuming a task in a cluster.

The way we can support using different secrets or workload identity would be by using kustomize to generate slightly different versions of the task.

tekton_configs/test-run.yaml
```diff
@@ -0,0 +1,58 @@
...
+   - name: xgboost-synthetic
+     taskRef:
+      name: pytest
+     params:
+     - name: src-dir
+       value: /mnt/testdir/src/gabrielwen
+     - name: repos
+       value: kubeflow/examples@HEAD,kubeflow/testing@HEAD,kubeflow/tf-operator@HEAD
+     - name: python-path
+       value: /mnt/testdir/src/gabrielwen/kubeflow/examples/py:/mnt/testdir/src/gabrielwen/kubeflow/testing/py:/mnt/testdir/src/gabrielwen/kubeflow/tf-operator/py
```
jlewi: Use Tekton git resources. So you should have a tekton resource for each repo. You should be able to reference the resources in setting the path e.g. something like

value: /workspace/$(inputs.resources.examples.name)/py:/workspace/$(inputs.resources.testing.name)/py

tekton_configs/test-run.yaml
```diff
@@ -0,0 +1,58 @@
...
+  pipelineSpec:
+   tasks:
+   - name: xgboost-synthetic
+     taskRef:
+      name: pytest
+     params:
+     - name: src-dir
+       value: /mnt/testdir/src/gabrielwen
+     - name: repos
+       value: kubeflow/examples@HEAD,kubeflow/testing@HEAD,kubeflow/tf-operator@HEAD
```
jlewi: Do we need tf-operator?

tekton_configs/test-run.yaml
```diff
@@ -0,0 +1,58 @@
...
+     - name: python-path
+       value: /mnt/testdir/src/gabrielwen/kubeflow/examples/py:/mnt/testdir/src/gabrielwen/kubeflow/testing/py:/mnt/testdir/src/gabrielwen/kubeflow/tf-operator/py
+     - name: get-cred-args-array
+       value:
+       - "-m"
+       - "kubeflow.testing.get_kf_testing_cluster"
+       - "get-credentials"
+     - name: pytest-args-array
+       value:
+        - xgboost_test.py
```
jlewi: Can we reduce the number of parameters that need to be supplied for the task?

tekton_configs/test-run.yaml
```diff
@@ -0,0 +1,58 @@
...
+        - --log-cli-level=info
+        - --log-cli-format='%(levelname)s|%(asctime)s|%(pathname)s|%(lineno)d| %(message)s'
+        - --timeout=1800
+        - --junitxml=/mnt/testdir/output/gabrielwen/junit_xgboost-synthetic-test.xml
+        - --notebook_artifacts_dir=/mnt/testdir/output/gabrielwen/xgboost-synthetic-test-notebooks
+        - --repos=kubeflow/examples@HEAD,kubeflow/tf-operator@HEAD
```
jlewi: It looks like the test currently launches a K8s job which will use init containers to pull down the repos as well as install papermill.

Should we add a step to this task that builds a container that bakes in the source code and papermill? Then the job we launch just needs to run papermill.

tekton_configs/pytest-task.yaml
gabrielwen: yea, adding test-notebook task is definitely fine.  but I think having a pytest task would be great because we have many tests under pytests and we plan to migrate other tests to pytest as well, right?

tekton_configs/pytest-task.yaml
gabrielwen: git resource is not able to change at runtime.  it's a PipelineResource that is similar to Task - we need to deploy with `kubectl apply` first.

tekton_configs/pytest-task.yaml
gabrielwen: same as above, git resource is not able to change revision at runtime.

tekton_configs/pytest-task.yaml
gabrielwen: since this is a generic pytest task, I'm thinking we might need different commands - for notebook test we use get_kf_cluster but for tests in kfctl we need to get the credential from the testing cluster.

tekton_configs/pytest-task.yaml
gabrielwen: yea we could - I found out that later and this is kept just because it's easier than making all the changes to directory paths.

tekton_configs/pytest-task.yaml
gabrielwen: we could if we don't need to provide the revision at runtime.

tekton_configs/test-run.yaml
gabrielwen: I'm not really sure - we can remove this and try when we start implementing things.

tekton_configs/test-run.yaml
gabrielwen: yes we definitely should.  this is a draft so just wanna give people a taste of this.

tekton_configs/test-run.yaml
gabrielwen: yea I think we definitely need to update part of our codes for the new convention, we can discuss this later.

jlewi:
I think the final PR might: kubeflow/testing/pull/622

PR #782, opened 2020-03-31
Remove multiple import statements in Mnist example

kurianbenoy:
In Kubeflow examples notebook, I found multiple import statements in kubeflow Mnist GCP deployment

### DIFF
--- a/mnist/mnist_gcp.ipynb
+++ b/mnist/mnist_gcp.ipynb
@@ -144,7 +144,6 @@
     }
    ],
    "source": [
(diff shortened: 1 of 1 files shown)


kurianbenoy:
@googlebot I signed it

NikeNano:
The jobs(xgboost-synthetic, mnist) seems to timeout for : 
```Job has reached the specified backoff limit, reason: BackoffLimitExceeded,
   status: 'True', type: Failed``` :(

jlewi:
I think the tests will continue to fail until #795 is commited.

PR #788, opened 2020-04-09
Fixes. End-to-End Pipeline Example on Azure

eedorenko:
- Fixed Score.py. Function run(raw_data) should accept a single argument json object
- Linting


berndverst:
@eedorenko why did you introduce a global variable and change the `run()` method signature in score.py?

I don't see a reason for this and I find that more confusing than what we had before.

eedorenko:
The thing is that function run(raw_data) in score.py must accept a single parameter - json object as it is described [here](https://microsoft.sharepoint.com/:w:/t/TEDDevOps/EXNbcuv4wu5NlXVNhLcMKMEBFn-e4atvMnyLp9VRRlt1KA?e=TGKrSH) and [here](https://microsoft.sharepoint.com/:w:/t/TEDDevOps/EXNbcuv4wu5NlXVNhLcMKMEBFn-e4atvMnyLp9VRRlt1KA?e=TGKrSH). Otherwise it will show an error "run() missing 1 required positional argument" while consuming the model web service as it is described [here](https://github.com/MicrosoftDocs/azure-docs/issues/16485)

berndverst:
Makes sense @eedorenko. Works for me!

berndverst:
/approve

Looks good from Azure perspective.

berndverst:
@jlewi could you merge this for us?

Looks huge because it's fixing linting errors, but there is just a tiny bit of refactoring in `score.py`.

/assign jlewi

jlewi:
/lgtm
/approve
/ok-to-test

@berndverst @eedorenko Can you add an OWNERs file to azurepipeline directory or list appropriate folks from Azure as approvers so you won't be blocked on me?

PR #791, opened 2020-04-20
Add OWNERS file for Azure Pipelines example

berndverst:
Adding Owners file for Azure example

Per suggestion from @jlewi in https://github.com/kubeflow/examples/pull/788#issuecomment-614167958


berndverst:
Blocked by https://github.com/kubeflow/internal-acls/pull/239

PR #823, opened 2020-09-22
update kfp 'github issue summarization' example

amygdala:
various changes to keep things up to date; update example to work with 'hosted KFP'.  Will create variant that works with a Kubeflow installation as a second phase.

(In case you follow the link to the accompanying codelab, it still needs to be updated too-- I'll do that today)


amygdala:
/assign @jlewi 
Jeremy, do you have time for this?

amygdala:
I don't believe the test errors would be related to my changes.

jinchihe:
```
No cluster named 'kf-vmaster-n00' in kubeflow-ci.
```

The test evn is changed, I think we need to enhance the CI test firstly or together. Thanks

amygdala:
> ```
> No cluster named 'kf-vmaster-n00' in kubeflow-ci.
> ```
> 
> The test evn is changed, I think we need to enhance the CI test firstly or together. Thanks

(I wasn't involved in setting up the tests and don't have a good grasp of how they work.) 

Perhaps it would be okay to pull in these changes regardless, which wouldn't have test coverage anyway.

amygdala:
Following up: I wonder if I could go ahead and pull this in (since test failures are unrelated), as it will update a broken codelab. 
If need be I could point the codelab to my fork instead of this repo, but I'd rather not.

PR #830, opened 2020-11-03
Removed form action attribute

senthilsivanath:
While accessing mnist ui from GCP hosted end point, takes to root url, as action attribute performs submit on /

Mnist ui url:
https:/${KUBEFlOW_ENDPOINT}/mnist/${NAMESPACE}/ui/   

Have removed action attribute, page will do submit on self.

Ref :
https://stackoverflow.com/questions/13520127/submit-html-form-on-self-page


senthilsivanath:
@googlebot I signed it!

jlewi (approved):

PR #831, opened 2020-11-06
pin version of h5py to 2.10.0 for Azure Pipeline Example

viraja1:
pin version of h5py to 2.10.0 since tensorflow 2.0.0 did not pin the version and h5py 3.0.0 broke the model loading code

#### Exception
'str' object has no attribute 'decode' for https://github.com/kubeflow/examples/blob/master/pipelines/azurepipeline/code/deploy/score.py#L21

https://github.com/tensorflow/tensorflow/issues/44467

### DIFF
--- a/pipelines/azurepipeline/code/deploy/environment.yml
+++ b/pipelines/azurepipeline/code/deploy/environment.yml
@@ -17,3 +17,4 @@ dependencies:
   - tensorflow==2.0.0-alpha0
   - Pillow
   - requests
+  - h5py==2.10.0
--- a/pipelines/azurepipeline/code/training/requirements.txt
+++ b/pipelines/azurepipeline/code/training/requirements.txt
@@ -1,2 +1,3 @@
 Pillow
 pathlib2
+h5py==2.10.0


pipelines/azurepipeline/code/deploy/environment.yml
```diff
@@ -16,4 +16,5 @@ dependencies:
   - numpy
   - tensorflow==2.0.0-alpha0
```
berndverst, line 17: Tensorflow 2 isn't Alpha anymore. How about instead we just change this line to:
`- tensorflow~=2`

pipelines/azurepipeline/code/deploy/environment.yml
```diff
@@ -16,4 +16,5 @@ dependencies:
...
   - Pillow
-  - requests
\ No newline at end of file
+  - requests
+  - h5py==2.10.0
```
berndverst, line 20: Update tensorflow library instead

pipelines/azurepipeline/code/training/requirements.txt
```diff
@@ -1,2 +1,3 @@
 Pillow
-pathlib2
\ No newline at end of file
+pathlib2
+h5py==2.10.0
```
berndverst, line 3: Probably not necessary?

berndverst (changes requested):

PR #876, opened 2021-08-08
Chinese-multiperson-voice-recognition-using-transfer-learning-main

sports75482:


sports75482:
@jinchihe  I have compressed the file, please check my file, thank you.

jinchihe:
@sports75482 Thanks the contribution.

1. Please create a directory for the example, instead of put them under root directory.
2. Why changed the License file? I think we should keep that.
3. Seems the `Chinese-multiperson-voice-recognition-using-transfer-learning-main.rar` is useless after unpacked to directly? if so, please remove that, thanks.

PR #917, opened 2022-02-01
Add recurring run demo

droctothorpe:
This PR demonstrates how to provision a [recurring run](https://www.kubeflow.org/docs/components/pipelines/concepts/run/) via the KFP SDK.


droctothorpe:
I elected to use a script instead of a notebook, but if notebooks are a firm requirement, I don't mind converting the script into a notebook.

js-ts:
thanks @droctothorpe for contributing and congratulations on your first PR

PR #1001, opened 2022-10-11
Transfer_learning Template

bebe424:


bebe424:
@js-ts This comitted code is a kubeflow template specifically for transfer learning applications.  It uses the spectrum image from Chinese speech as input data, fed it into the hidden layers transfered from a pre-trained model, e.g. Vgg16.   The complete process is targeted toward rapid replacement of the pre-trained model parameters.  Please feel free to ask any questions you might have.  We hope to contribute to this community. thanks

PR #1008, opened 2022-11-18
CNN template

ifuwang:


ifuwang (commented):
@kimwnasptd 
Dear Sir, we are committing this template code to be used by users in this community for Kubeflow pipeline reuse via a template construct.  This example takes a mnist CNN training process as illustration.  Please help review the code and make it recognized.

PR #1016, opened 2022-11-22
Update the examples with correct image paths and packages

kimwnasptd:
Some of the examples had issues with
1. The images would not get rendered correctly, due to jupyter not parsing query params correctly
2. The Kale image was explicitly set in the notebook metadata
3. Some packages needed some updates

cc @akgraner @jimmyguerrero


akgraner:
Thank you for your contribution!
/lgtm
/approve

PR #1035, opened 2023-05-05
Update demo of mnist on kfp version 1.8.19

kbthu:


kbthu:
This example is verified on KFP 1.8.19 which will help to show a training case for mnist model.
cc  @akgraner

PR #1067, opened 2023-09-21
Nodered integrate Kubeflow with IBMQ Qiskit example

justin0322:
The goal of node-red-quantum-kube is to integrate Kubeflow/Kubebeters with Node-RED, leveraging Node-RED's low-code modules and utilizing Kubeflow resources, such as Kubeflow pipelines, to enhance its artificial intelligence and machine learning capabilities. As part of our demonstration, we also provide a practical example illustrating the use of Kubeflow. Specifically, we showcase how to leverage a hybrid quantum-classical neural network using TorchConnector to perform a more complex image classification task on the MNIST handwritten digits dataset."


kbthu (commented):
I have read your codes and readme files.  This is an interesting live example of integrating kubeflow with quantum machine learning.  More, it also features an integration with low code GUI of Node-red to exercise the quantum ML pipeline.  However the introduction part of readme file needs more explanation of the usage of the whole package.  This will allow readers and users of this code to understand the complete procedures of applying to general purpose applications.

kbthu (approved):
After viewing the changed readme file, it is clear that users can easily reproduce the quantum neural network kubeflow pipeline embedded under the Node-red front end.  With this feature, users can get access to the computing power of a quantum machine with only a few move of the a mouse.  I will happy to give it a go.

Issue #1109, opened 2024-05-18
Revamping the examples repository


dpoulopoulos:
The `kubeflow/examples` repository is a valuable resource for demonstrating how to use Kubeflow for machine learning. However, it has fallen out of date, with examples using outdated libraries and lacking maintenance. I'd like to suggest a list of actions to address these issues and improve the overall quality and relevance of this repository:

1. Restructure Repository Content
    * **End-to-End Examples Only**: Retain only the end-to-end examples in this repository.
    * **Tool-Specific Examples**: Move framework-specific examples to their corresponding repositories, where their respective working groups will be responsible for maintenance.
1. Code Overhaul
    * **Review and update existing examples**: Update the code to use the latest versions of libraries and frameworks, ensuring compatibility with the current Kubeflow version.
    * **Develop and integrate new examples**: Identify and prioritize new examples that demonstrate cutting-edge use cases and advanced features of Kubeflow.
1. Branching Model
    * **Introduce Development Branch**: Establish a `develop` branch (the default branch) for ongoing development work.
    * **Release Management**: Create specific release branches from the develop branch that follow Kubeflow releases.
1. Best Practices and Repository Management (for example)
    * **Main branch Protection**: Protect the main repository to ensure stability.
    * **PR Reviews and Merging**: Enforce the practice of PR reviews before merging changes.
    * **Templates for PRs and Issues**: Introduce standardized templates for creating Pull Requests and Issues to streamline contributions.
1. Documentation Updates
    * **README**: Update the README file to reflect the changes and provide clear instructions on how to use and contribute to the repository.
    * **CONTRIBUTING**: Revise the CONTRIBUTING file to include the new contribution guidelines and best practices.

If you agree with the proposal, I'd like to get involved. Based on this suggestion, let's start by identifying and categorizing existing examples into end-to-end and tool-specific. Then, we can create and configure a develop branch that will house the end-to-end examples and begin introducing the code and documentation changes we need.

StefanoFioravanzo:
@dpoulopoulos Thanks for starting this issue. I couldn't agree more with all the action items you raised! Is there a particular sequenced priority that you would like to recommend?

cc some people who might be interested in this @andreyvelich @chasecadet @rimolive

kimwnasptd:
Really nice proposal!

[...]

PR #1178, opened 2024-11-28
Notify users that these examples are out-of-date

andreyvelich:
As we discussed on [the latest Kubeflow community call](https://docs.google.com/document/d/1Wdxt1xedAj7qF_Rjmxy1R0NRdfv7UWs-r2PItewxHpE/edit?tab=t.0), we want to notify users that some examples from this repo are broken.
We got feedback that many new users get started with this repo and can't run the Kubeflow examples.

We don't want to archive this repo for now, since we might want to give this repo second live in the future: https://github.com/kubeflow/examples/issues/1109

cc @kubeflow/kubeflow-steering-committee @kubeflow/wg-data-leads @kubeflow/wg-pipeline-leads @kubeflow/wg-notebooks-leads @kubeflow/wg-training-leads @kubeflow/wg-automl-leads @kubeflow/wg-manifests-leads


juliusvonkohout:
I would link to the community page instead of the Google document but otherwise lgtm

README.md
```diff
@@ -1,12 +1,23 @@
-## Notice
+# :warning: **kubeflow/examples Repository is not Maintained**
```
tarilabs, line 1: TIL, I was [aware of admonitions](https://github.com/orgs/community/discussions/16925#discussion-4085374) but not this option :)

